PR #4, opened 2018-05-25
Initial NEMU CI

sameo:
This is the first basic `semaphoreci` and `pullapprove` based CI for NEMU. We have:

* A `pullapprove` configuration file where 2 LGTMs from the nemu-write team are needed for a PR to be merged.
* A couple of scripts for running minimal CI in a `semaphoreci` context

### DIFF
--- a/pc-bios/optionrom/code16gcc.h
+++ b/pc-bios/optionrom/code16gcc.h
@@ -1,0 +1,3 @@
+asm(
+".code16gcc\n"
+);
--- a/tools/CI/minimal_ci.sh
+++ b/tools/CI/minimal_ci.sh
@@ -31,7 +31,6 @@ CLOUD_INIT_DIR="$SCRIPT_DIR/cloud-init/"
 BUILD_DIR=""
 HYPERVISOR="$HOME/build-x86_64/x86_64-softmmu/qemu-system-x86_64"
 WORKLOADS_DIR="$HOME/workloads"
-EFI_FIRMWARE="$HOME/workloads/OVMF.fd"
 EFI_FIRMWARE_URI="https://download.clearlinux.org/image/OVMF.fd"
 DOWNLOAD_ONLY="false"
 RUN_UNSAFE="false"
@@ -234,6 +233,7 @@ if [ ! -d "$WORKLOADS_DIR" ]; then
   fi
 fi
 
+EFI_FIRMWARE="$WORKLOADS_DIR/OVMF.fd"
 if [ ! -f $EFI_FIRMWARE ]; then
    wget --no-verbose -O "$EFI_FIRMWARE" "$EFI_FIRMWARE_URI"
 fi
--- a/tools/CI/minimal_ci.sh
+++ b/tools/CI/minimal_ci.sh
@@ -66,7 +66,7 @@ run_tests() {
 
        if [ ! -f "$WORKLOADS_DIR"/"$image" ]; then
           pushd "$WORKLOADS_DIR"
-          wget "$uri"
+          wget --no-verbose "$uri"
           popd
        fi
    
@@ -235,7 +235,7 @@ if [ ! -d "$WORKLOADS_DIR" ]; then
 fi
 
 if [ ! -f $EFI_FIRMWARE ]; then
-   wget -O "$EFI_FIRMWARE" "$EFI_FIRMWARE_URI"
+   wget --no-verbose -O "$EFI_FIRMWARE" "$EFI_FIRMWARE_URI"
 fi
 
 IS_AARCH64=`file $HYPERVISOR | grep aarch64`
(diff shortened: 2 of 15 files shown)


sameo:
@mcastelino @rbradford Please not that we need to get additional privileges in order to enable pullapprove and semaphore CI to gate our wip-master branch.

pc-bios/optionrom/code16gcc.h
```diff
@@ -0,0 +1,3 @@
+asm(
```
mcastelino: @sameo why was this needed?

tools/CI/minimal_ci.sh
```diff
@@ -37,6 +37,7 @@ DOWNLOAD_ONLY="false"
 RUN_UNSAFE="false"
 VERBOSE="false"
 CHECK="false"
+VSOCK="true"
```
mcastelino: thanks. I should have done this before. This will also allow folks to test on older host kernels.

pc-bios/optionrom/code16gcc.h
sameo: The semaphoreci toolchain is gcc 4.8, and  from pc-bios/optionrom/Makefile:

```
ifeq ($(filter -m16, $(QEMU_CFLAGS)),)
# Attempt to work around compilers that lack -m16 (GCC <= 4.8, clang <= ??)
# On GCC we add -fno-toplevel-reorder to keep the order of asm blocks with
# respect to the rest of the code.  clang does not have -fno-toplevel-reorder,
# but it places all asm blocks at the beginning and we're relying on it for
# the option ROM header.  So just force clang not to use the integrated
# assembler, which doesn't support .code16gcc.
QEMU_CFLAGS += $(call cc-option, $(QEMU_CFLAGS), -fno-toplevel-reorder)
QEMU_CFLAGS += $(call cc-option, $(QEMU_CFLAGS), -no-integrated-as)
QEMU_CFLAGS += -m32 -include $(SRC_PATH)/pc-bios/optionrom/code16gcc.h
endif
```

sameo:
@mcastelino It highly depends on the networking bandwidth, but with a poor one it can take up to 1h.
I will have a look at enabling generic cache for caching the workloads images across builds.
If that does not work, we'll have to strip it down to 2-3 workloads, yes.

tools/CI/semaphore/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+# -fno-semantic-interposition is not supported by gcc 4.8 (from semaphore)
+export EXTRA_CFLAGS=" -O3 -falign-functions=32 -D_FORTIFY_SOURCE=2 -fPIE"
+export SRCDIR=~/nemu
+
+sudo apt-get install -y libcap-ng-dev librbd-dev
+
+$SRCDIR/tools/build_x86_64.sh
```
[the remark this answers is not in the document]
sameo: > It would also allow the testing of vsock as the ccloudvm scripts use bionic which runs with a kernel new enough to do vsock.

Are you suggesting to also run the actual CI scripts inside ccloudvm?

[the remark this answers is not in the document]
rbradford: @sameo I run the CI inside ccloudvm myself. We've successfully run qemu inside ccloudvm inside semaphore for CIAO so I think it should work.

tools/CI/semaphore/setup.sh
```diff
@@ -0,0 +1,38 @@
...
+sudo apt-get update
+sudo apt-get install qemu xorriso -y
+sudo chmod ugo+rwx /dev/kvm
+
+echo ""
+echo "===== Building ccloudvm ====="
+echo ""
+
+go version
+go get github.com/intel/ccloudvm
```
rbradford: @sameo if you do go get github.com/intel/ccloudvm/... then it will do the right thing with the packages removing the need for the line below.

rbradford:
@sameo with the changes i've pushed building and cross building works

I've also also pushed a change to run the CI script inside the ccloudvm, let's see if that works.

sameo:
@rbradford Thanks for fixing the ccloudvm based build.
It seems running the CI inside ccloudvm failed.
I cleaned up the branch (I had a debug commit there) and removed your last commit to double check if CI passes now:

https://semaphoreci.com/sameo/nemu/branches/topic-ci/builds/38

rbradford:
@sameo I've fixed the PATH inside the script now so it should hopefully find ccloudvm to do the run command.

Irrespective of that it seems to timeout after 1 hour mostly from downloading all the files. Either we could reduce what we do (based off some of our recently discoveries) or maybe use semaphore caching: https://semaphoreci.com/docs/caching-between-builds.html

sameo:
@rbradford Thanks a lot, LGTM.

PR #18, opened 2018-07-02
[WIP] Direct kernel boot for virt/i386

### LINKED ISSUE #13: Implement direct kernel boot for i386/virt
Since the i386 virt machine type presents a minimal hardware list, we should be able to do direct kernel boot without much complications.
This should be an opt-in choice, same as the `nofw` option from `pc-lite`.

sameo:
In order to decouple the virt/i386 machine type implementation from the OVMF dependencies, we want to add support for a `nofw` machine type acceleration option. `nofw` skips the firmware loading and booting step and does a direct kernel boot.

- [x] Factorize all ACPI tables build code across i386 machine types
- [x] Implement fw_cfg support for i386/virt
- [x] Build ACPI tables for i386/virt
- [x] Backport pc-lite's nofw support

Fixes #13 


yangzhon (commented):
hw/i386/pc.c

+    /* Machine class settings */
+    conf->legacy_acpi_table_size = pcmc->legacy_acpi_table_size;
+    conf->legacy_cpu_hotplug = pcmc->legacy_cpu_hotplug;
+    conf->rsdp_in_ram = pcmc->rsdp_in_ram;
+
+    /* Machine state settings */
+    conf->fw_cfg = pcms->fw_cfg;
+    conf->acpi_dev = pcms->acpi_dev;
+    conf->below_4g_mem_size = pcms->below_4g_mem_size;
+    conf->numa_nodes = pcms->numa_nodes;
+    conf->node_mem = pcms->node_mem;
+    conf->apic_xrupt_override = pcms->apic_xrupt_override;
+    conf->apic_id_limit = pcms->apic_id_limit;
+    conf->acpi_nvdimm_state = pcms->acpi_nvdimm_state;
+    conf->hotplug_memory = pcms->hotplug_memory;

In the virt machine type, there is not related code to set pcms->below_4g_mem_size in virt_memory_init(), right?
same for other variants in pcms.

liujing2:
below_4g_mem_size is just 3G, hard code? And VirtMachineState has a bool field called "highmem" which is not used?

yangzhon (commented):
If we create one new file to include fw functions, whether the pc_build_smbios() is also in this file ?  thanks!

sameo:
@yangzhon 
> If we create one new file to include fw functions, whether the pc_build_smbios() is also in this file ? thanks!

Yes, we should do that. I'll amend the fw.c file.

sameo:
@yangzhon @liujing2 
> same for other variants in pcms.

Yes. There are some missing fields needed to build the ACPI tables. Since a many of them are in the pc machine type for legacy reasons, I'm adding them only when needed. We're not going to support legacy platforms.

sameo:
> > @yangzhon
> >  If we create one new file to include fw functions, whether the pc_build_smbios() is also in this file ? thanks!

> Yes, we should do that. I'll amend the fw.c file.

See 69f67df430aa7873a8a8749b86274c76d9d085a4

liujing2 (commented):
Samuel, I didn't quite understand why we replace pcms to acpiconf. PCMachineState is also a common definition and the AcpiConfiguration is almost the same copy of pcms? Does this mean we want to totally remove pcms?

sameo:
@liujing2 
> Samuel, I didn't quite understand why we replace pcms to acpiconf.

`pcms` and `pcmc` are really piix and q35 related and we want the i386 code base to be as machine agnostic as possible. One should be able to define a new i386 machine type without carrying all the piix legacy for example.

Yes, this means there is some duplication between pcms and AcpiConf, but that's the best compromise I could come up with for decoupling the ACPI build APIs from the piix and q35 machine types.

rbradford:
I've looked through this today. I want to take a closer look at it a bit more tomorrow. But here are my initial thoughts:

* With the refactoring work and also with the changes to the direct booting have we been testing that Q35 and PIIX are working fine with direct boot and not?
* Some of the commit messages could do with a bit of elaboration. In particular some of them just say what is done but not why.
* virt.c is pretty short which is a great sign that this is heading in the right direction
* I want to give this a build and test run myself.

I'll continue looking more closely tomorrow after building and running it but overall looks pretty good.

sameo:
@rbradford I fixed some of the light commit messages, and I'm going to give this ago with some cloud images on top of pc and q35.

sameo:
@rbradford @yangzhon Manually running minimal-ci.sh with this PR passes. This is ready to merge.

PR #20, opened 2018-07-12
CPU, PCI and firmware bits

rbradford:
@chao-p do my changes to pci_lite.c look okay? Mainly it was changing the way the hole details were stored and accessed.


rbradford:
@sameo here is a bit of a "mega" PR with my work so far:

* I add the basic support for CPU hotplug so that the datastructures get populated enough (fix segfault on non-direct boot)
* Refactor and use the firmware loading code (OVMF firmware is now loaded.)
* Import and use the PCI lite host bridge (OVMF sees the PCI host bridge)

sameo:
@rbradford This PR LGTM, but I'd like to get @yangzhon review on the pcie-lite backport before merging.

yangzhon:
@rbradford, @sameo

This PR LGTM, but there are two issues we need care.
(1). Since we support PCIe bus, so the legacy IRQ setting are not useful now. We will remove this.
(2). As for cpu hotplug, your patches can plug the cpus as qemu command settings. In PC and Q35 platform, which also support ACPI to hotplug cpus by qemu monitor, this need gpe event and SCI interrupt supported in acpi table. We plan to use reduced HW acpi, so, how can we deal with cpu and memory hotplug?

sameo (approved):

PR #21, opened 2018-07-17
tools: Add master branch CI back in

sameo:
We need it to properly run our semaphore based CI.
This could partially be upstreamed as well.


yangzhon (commented):
This PR LGTM.

btw, for PCIe hotplug, we should use pcie-root-port to replace pci-bridge. We will replace it in next step. thanks!

PR #22, opened 2018-07-17
Inital ACPI reduced implementation for i386/virt

sameo:
We build a minimal set of ACPI hw reduced tables: RSDT, FADT, DSDT pointed by a RSDP.
The DSDT only contains one PCI host bridge.

This will be directly consumed by the i386 virt platform.


rbradford:
Looks good. What are the planned next steps. I see that the cold boot SMP vCPUs don't appear and we still have something to do to setup the PCI bus correctly. But the kernel definitely seems happier with these tables now!

rbradford (approved):

yangzhon (approved):
@sameo , this PR LGTM, thanks!

sameo:
@rbradford Next steps are about adding the MADT table and the PCI irq routing. The former should fix the SMP issue.

PR #24, opened 2018-07-19
MADT table addition

sameo:
The MADT ACPI table describes the platform interrupt controllers and the guest kernels builds the SMP information from it.

As the i386 MADT ACPI build routine depends on a system ACPI device, we first add a (for now) dummy ACPI device hanging of the system bus. Then we safely add the MADT tables to the set of tables we build for the virt i386 machine type.

### DIFF
--- a/hw/i386/virt/acpi.c
+++ b/hw/i386/virt/acpi.c
@@ -1,0 +1,123 @@
+/*
+ *
+ * Copyright (c) 2018 Intel Corporation
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms and conditions of the GNU General Public License,
+ * version 2 or later, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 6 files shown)


hw/i386/virt/acpi.c
```diff
@@ -0,0 +1,123 @@
...
+static void virt_send_ged(AcpiDeviceIf *adev, AcpiEventStatusBits ev)
+{
+}
+
+static int virt_device_sysbus_init(SysBusDevice *dev)
+{
+    return 0;
+}
+
+DeviceState *virt_pm_init(void)
```
rbradford: @sameo Can I be a pain and suggest we don't call it virt_pm but virt_acpi/VIRT_ACPI/VirtAcpi/etc, we're not going to be doing any of the power management bits of ACPI so using "PM" feels wrong...

rbradford (approved):
Nomenclature change *recommendation*. Functionality lgtm.

yangzhon (commented):
Since the reduced HW acpi do not support PM control any more in virt, do we still need PM control ? thanks!

hw/i386/virt/acpi.c
sameo: Done.

yangzhon:
LGTM

PR #25, opened 2018-07-23
hw: acpi: Implement XSDT support for RSDP

sameo:
XSDT is the 64-bit version of the legacy ACPI RSDT (Root System
    Description Table). RSDT only allow for 32-bit addressses and have thus
    been deprecated. Since ACPI version 2.0, RSDPs should point at XSDTs and
    no longer RSDTs, although RSDTs are still supported for backward
    compatibility.

Since version 2.0, RSDPs should add an extended checksum, a complete table
length and a version field to the table.
    

yangzhon (commented):
LGTM.

@sameo, you can merge this PR before my patches. and my patches should be ready tomorrow. thanks!

PR #26, opened 2018-07-23
add pci dsdt and mmcfg support

yangzhon:
Those patches added the pci dsdt support from q35/pc platform and guest kernel can normally show pci bus log.  Since virt/i386 default bus is pcie, i added the mmcfg support for guest kernel.

Some issues need to be talked
(1) build_prt(), build_crs(), build_gsi_link_dev(), build_osc_method() are moved to  hw/acpi/aml-build.c ? 
     The current patch is still in i386/acpi-build.c

Please @sameo @rbradford @liujing2 help review those 3 patches, thanks!


include/hw/acpi/aml-build.h
```diff
@@ -422,4 +422,28 @@ void build_slit(GArray *table_data, BIOSLinker *linker);
 
 void build_fadt(GArray *tbl, BIOSLinker *linker, const AcpiFadtData *f,
                 const char *oem_id, const char *oem_table_id);
+
+/*acpi-build.c*/
```
sameo: Nitpick: Either we add a comment that describes what those structures are, and not simply where they come from, or we don't comment at all. I prefer the former.

include/hw/acpi/aml-build.h
```diff
@@ -422,4 +422,28 @@ void build_slit(GArray *table_data, BIOSLinker *linker);
...
+
+Aml *build_prt(bool is_pci0_prt);
+Aml *build_gsi_link_dev(const char *name, uint8_t uid, uint8_t gsi);
+Aml *build_osc_method(void);
+void crs_range_set_init(CrsRangeSet *range_set);
+Aml *build_crs(PCIHostState *host, CrsRangeSet *range_set);
+void crs_replace_with_free_ranges(GPtrArray *ranges,
+                                  uint64_t start, uint64_t end);
+
+void crs_range_set_free(CrsRangeSet *range_set);
```
sameo: Another nit: Can we group the `crs_*` routines definitions together?

hw/i386/acpi-build.c
```diff
@@ -1727,7 +1722,7 @@ static void build_piix4_pci_hotplug(Aml *table)
     aml_append(table, scope);
 }
 
-static Aml *build_q35_osc_method(void)
+Aml *build_osc_method(void)
```
sameo: All those general routines are not x86 specific, so could you please create a separate patch that move them from `i386/acpi-build.c` to `aml-build.c` ?

hw/i386/acpi-build.c
```diff
@@ -255,10 +255,16 @@ static Object *acpi_get_i386_pci_host(void)
                             TYPE_PCI_HOST_BRIDGE);
     }
 
+    if (!host) {
+        host = OBJECT_CHECK(PCIHostState,
+                            object_resolve_path("/machine/pcilite", NULL),
+                            TYPE_PCI_HOST_BRIDGE);
+    }
+
```
sameo: That one is a `acpi_get_i386_pci_host` fix, and I'd love to see it as a standalone patch in this PR.
I also think that it should not be i386 specific and we could have a list of QEMU supported PCI host strings list and have this routine look like that:

```C
static char *pci_hosts[] = {
   "/machine/i440fx",
   "/machine/q35",
   "/machine/pcilite",
   NULL,
};

static Object *acpi_get_pci_host(void)
{
    PCIHostState *host;
    int i = 0;

    while (pci_hosts[i]) {
          host = OBJECT_CHECK(PCIHostState,
                        object_resolve_path(pci_hosts[i], NULL),
                        TYPE_PCI_HOST_BRIDGE);
          if (host) {
              return OBJECT(host);
          }

          i++;
    }

    return NULL;
}
```

hw/acpi/reduced.c
```diff
@@ -55,10 +57,31 @@ static void acpi_dsdt_add_cpus(Aml *scope, int smp_cpus)
...
+
+    aml_append(pci0_scope, build_prt(false));
+    aml_append(sb_scope, pci0_scope);
+
+    aml_append(sb_scope, build_gsi_link_dev("LNKA", 0x10, 0x10));
+    aml_append(sb_scope, build_gsi_link_dev("LNKB", 0x11, 0x11));
+    aml_append(sb_scope, build_gsi_link_dev("LNKC", 0x12, 0x12));
+    aml_append(sb_scope, build_gsi_link_dev("LNKD", 0x13, 0x13));
+    aml_append(table, sb_scope);
+}
```
sameo: I'd prefer for this one to not be under `acpi/reduced.c`. Is it generic enough to make it part of `aml-build.c` ?

hw/acpi/reduced.c
```diff
@@ -77,9 +100,110 @@ static void build_dsdt(GArray *table_data, BIOSLinker *linker)
...
+    aml_append(dev, aml_name_decl("SUPP", aml_int(0)));
+    aml_append(dev, aml_name_decl("CTRL", aml_int(0)));
+    aml_append(dev, build_osc_method());
     aml_append(scope, dev);
     aml_append(dsdt, scope);
 
+    build_lite_pci0_int(dsdt);
+    crs_range_set_init(&crs_range_set);
+    bus = VIRT_MACHINE(machine)->pci_bus;
+    if (bus) {
```
sameo: We should assert that there is a bus. We won't support cases where `bus` is NULL.

hw/acpi/reduced.c
```diff
@@ -77,9 +100,110 @@ static void build_dsdt(GArray *table_data, BIOSLinker *linker)
...
+                                        0, entry->limit - entry->base + 1));
+        }
+    }
+
+    aml_append(scope, aml_name_decl("_CRS", crs));
+
+    crs_range_set_free(&crs_range_set);
+
+    aml_append(dsdt, scope);
+
```
sameo: A few comments on this chunk:

* There are quite a few hardcoded numbers: `0x0CF8, 0x0CF8, 0x01, 0x080x000A0000, 0x000BFFFF, 0, 0x00020000`, etc. We need to have definitions for those.
* There are no comments, we need to put some explanations on what we're trying to do here.
* `build_dsdt` is now getting too big. We should split it by e.g. putting the above block of code into its own separate function, for PCI DSDT settings.

hw/acpi/reduced.c
```diff
@@ -257,6 +258,10 @@ static void acpi_reduced_build(MachineState *ms, AcpiBuildTables *tables, AcpiCo
     acpi_add_table(table_offsets, tables_blob);
     mc->firmware_build_methods.acpi.madt(tables_blob, tables->linker, ms, conf);
 
+    if (acpi_get_mcfg(&mcfg)) {
+        acpi_add_table(table_offsets, tables_blob);
+        build_mcfg(tables_blob, tables->linker, &mcfg);
```
sameo: Could we please add a `mcfg` method to `struct FirmwareBuildMethods`, set `i386/virt.c` `mcfg` pointer to point at `build_mcfg` and call `mc->firmware_build_methods.acpi.mcfg`.

hw/i386/acpi-build.c
```diff
@@ -2432,8 +2427,8 @@ build_srat(GArray *table_data, BIOSLinker *linker, MachineState *machine, AcpiCo
                  table_data->len - srat_start, 1, NULL, NULL);
 }
 
-static void
-build_mcfg_q35(GArray *table_data, BIOSLinker *linker, AcpiMcfgInfo *info)
+void
+build_mcfg(GArray *table_data, BIOSLinker *linker, AcpiMcfgInfo *info)
```
sameo: Could we please move that one to `aml-build.c` as well?

include/hw/acpi/acpi.h
```diff
@@ -183,6 +185,11 @@ void acpi_send_gpe_event(ACPIREGS *ar, qemu_irq irq,
 
 void acpi_update_sci(ACPIREGS *acpi_regs, qemu_irq irq);
 
+/*acpi-build.c*/
```
sameo: Same comment as before: It would be better to have a more meaningful comment here.

sameo (changes requested):

sameo:
@yangzhon Thanks for the PR. I have a few comments, mostly about code reorganization and factorization.

hw/i386/acpi-build.c
yangzhon: okay, got it, thanks!

hw/i386/acpi-build.c
yangzhon: Yes, done,thanks!

hw/i386/acpi-build.c
yangzhon: Yes, done, thanks!

include/hw/acpi/aml-build.h
yangzhon: @sameo , the below crs_* related struct and functions are used in general, other are static functions in hw/i386/acpi-build.c. What's youe meaning about "group" ?   remove the line 445 as above shown or move all crs_* here? thanks!

btw, whether we need move all crs_* related functions into aml-build.c ?

typedef struct CrsRangeEntry {                                                 
    uint64_t base;                                                             
    uint64_t limit;                                                            
} CrsRangeEntry;                                                               
                                                                               
typedef struct CrsRangeSet {                                                   
    GPtrArray *io_ranges;                                                      
    GPtrArray *mem_ranges;                                                     
    GPtrArray *mem_64bit_ranges;                                               
} CrsRangeSet;                                                                 
                                                                                                                         
void crs_range_set_init(CrsRangeSet *range_set);                               
Aml *build_crs(PCIHostState *host, CrsRangeSet *range_set);                    
void crs_replace_with_free_ranges(GPtrArray *ranges, uint64_t start, uint64_t end);               
void crs_range_set_free(CrsRangeSet *range_set);                               
void acpi_get_pci_holes(Range *hole, Range *hole64);

hw/acpi/reduced.c
yangzhon: Yes, remove it to aml-build.c, thanks!

hw/acpi/reduced.c
yangzhon: Yes, you are right, the assert(bus) will be added, thanks!

include/hw/acpi/aml-build.h
sameo: @yangzhon I just meant to put the `crs_*` functions next to each others in the header file, but I see that they're already are grouped together. Forget about my comment, sorry for the confusion.

I think it would make sense to move all the `crs_*` routines into aml-build.c if they're not i386 specific.

include/hw/acpi/aml-build.h
yangzhon: yes, i will add comments here, thanks!

hw/acpi/reduced.c
yangzhon: thanks, new version did it.

include/hw/acpi/acpi.h
yangzhon: The new version patches did not add definition here, all definitions are in aml-build.h, so omit this. thanks!

hw/acpi/reduced.c
yangzhon: Yes, i moved the pci related acpi function into aml-build.c and build_dsdt() only call  acpi_dsdt_add_pci() for pci acpi setting like arm did.

yangzhon: @sameo ,

I add the macro define for 0xcf8, 0xa0000 and etc, which make code more clear.
The comments are also added into the related code to explain which mem or io region is set in the aml table.  thanks!

yangzhon:
The new patches are ready, and i changed those according to samuel's comments. thanks sameo!
 
Please @sameo @rbradford @liujing2 help review those 5 patches, thanks!

hw/acpi/aml-build.c
```diff
@@ -1824,6 +1838,155 @@ Aml *build_prt(bool is_pci0_prt)
...
+{
+    Aml *sb_scope = aml_scope("_SB");
+    Aml *pci0_scope = aml_scope("PCI0");
+    aml_append(pci0_scope, build_prt(false));
+    aml_append(sb_scope, pci0_scope);
+
+    aml_append(sb_scope, build_gsi_link_dev("LNKA", 0x10, 0x10));
+    aml_append(sb_scope, build_gsi_link_dev("LNKB", 0x11, 0x11));
+    aml_append(sb_scope, build_gsi_link_dev("LNKC", 0x12, 0x12));
+    aml_append(sb_scope, build_gsi_link_dev("LNKD", 0x13, 0x13));
```
sameo: This PCI irq routing is quite standard and not really PCI lite specific, isn't it?
I would rename this one to something like `build_pci0_standard_int`.

hw/acpi/aml-build.c
```diff
@@ -1824,6 +1838,155 @@ Aml *build_prt(bool is_pci0_prt)
...
+{
+    CrsRangeEntry *entry;
+    Aml *scope, *dev, *crs;
+    CrsRangeSet crs_range_set;
+    PCIBus *bus = NULL;
+    int root_bus_limit = 0xFF;
+    int i;
+
+    crs_range_set_init(&crs_range_set);
+    bus = VIRT_MACHINE(machine)->pci_bus;
```
sameo: Please pass the `PCIBus` pointer as a function argument. We should not be linked to the `i386/virt` code in aml-build.c.

hw/acpi/aml-build.c
```diff
@@ -1824,6 +1838,155 @@ Aml *build_prt(bool is_pci0_prt)
...
+    aml_append(sb_scope, pci0_scope);
+
+    aml_append(sb_scope, build_gsi_link_dev("LNKA", 0x10, 0x10));
+    aml_append(sb_scope, build_gsi_link_dev("LNKB", 0x11, 0x11));
+    aml_append(sb_scope, build_gsi_link_dev("LNKC", 0x12, 0x12));
+    aml_append(sb_scope, build_gsi_link_dev("LNKD", 0x13, 0x13));
+    aml_append(table, sb_scope);
+}
+
+static void build_lite_pci_host_bridge(Aml *table, Range *pci_hole,
```
sameo: Same here: How is this PCI lite specific? Isn't that routine building a standard, minimal PCI host bridge AML?

hw/acpi/aml-build.c
```diff
@@ -25,6 +25,20 @@
...
+
+#define PCI_HOST_BRIDGE_CONFIG_ADDR        0xcf8
+#define PCI_HOST_BRIDGE_IO_0_MIN_ADDR      0x0000
+#define PCI_HOST_BRIDGE_IO_0_MAX_ADDR      0x0cf7
+#define PCI_HOST_BRIDGE_IO_1_MIN_ADDR      0x0d00
+#define PCI_HOST_BRIDGE_IO_1_MAX_ADDR      0xffff
+#define PCI_VGA_MEM_BASE_ADDR              0x000a0000
+#define PCI_VGA_MEM_MAX_ADDR               0x000bffff
+#define IO_0_LEN                           0xcf8
+#define VGA_MEM_LEN                        0x20000
```
sameo, line 44: Nice, thanks.

hw/acpi/aml-build.c
```diff
@@ -1824,6 +1838,155 @@ Aml *build_prt(bool is_pci0_prt)
...
+    aml_append(dev, aml_name_decl("_ADR", aml_int(0)));
+    aml_append(dev, aml_name_decl("_UID", aml_int(1)));
+    aml_append(dev, aml_name_decl("SUPP", aml_int(0)));
+    aml_append(dev, aml_name_decl("CTRL", aml_int(0)));
+    aml_append(dev, build_osc_method());
+    aml_append(sb_scope, dev);
+    aml_append(table, sb_scope);
+
+    build_lite_pci0_int(table);
+    build_lite_pci_host_bridge(table, pci_hole, pci_hole64, machine);
```
sameo: So this function should also take a `PCIBus` pointer as an argument, and if it is not `NULL` then we call the 2 above functions. This will make this part of the code not dependent on `i386/virt` and reusable if we ever decide to not have a PCI bus at all.

hw/acpi/reduced.c
```diff
@@ -56,9 +56,11 @@ static void acpi_dsdt_add_cpus(Aml *scope, int smp_cpus)
 }
 
 /* DSDT */
-static void build_dsdt(GArray *table_data, BIOSLinker *linker)
+static void build_dsdt(GArray *table_data, BIOSLinker *linker,
+                       Range *pci_hole, Range *pci_hole64,
+                       MachineState *machine)
```
sameo: Ditto here. Maybe we should even have a ACPI specific PCI structure:

```C
struct AcpiPciBus {
   PCIBus *pci_bus;
   Range *pci_hole;
   Range *pci_hole64;
};
```

Up to you.

include/hw/acpi/aml-build.h
```diff
@@ -231,6 +231,26 @@ typedef struct AcpiMcfgInfo {
...
+    GPtrArray *io_ranges;
+    GPtrArray *mem_ranges;
+    GPtrArray *mem_64bit_ranges;
+} CrsRangeSet;
+
+void crs_range_set_init(CrsRangeSet *range_set);
+Aml *build_crs(PCIHostState *host, CrsRangeSet *range_set);
+void crs_replace_with_free_ranges(GPtrArray *ranges,
+                                  uint64_t start, uint64_t end);
+void crs_range_set_free(CrsRangeSet *range_set);
```
sameo: So if we declare those functions in `aml-build.h` but define them in `i386/acpi-build.c` a non i386 machine type won't be able to include aml-build.h.
So I think we should define them in `aml-build.c`. @yangzhon Why is it difficult?

sameo (changes requested):

hw/acpi/reduced.c
yangzhon: Yes, i did it, thanks!

hw/acpi/aml-build.c
yangzhon: Yes, i changed this, thanks!

include/hw/acpi/aml-build.h
yangzhon: I have moved those routines into aml-build.c, thanks!

hw/acpi/aml-build.c
yangzhon: New code has changed it, thanks!

hw/acpi/aml-build.c
yangzhon: Yes, this build_lite_pci_host_bridge can be shared between virt-i386, pc and q35.

hw/acpi/aml-build.c
yangzhon: Yes, did it, thanks!

hw/acpi/aml-build.c
yangzhon: @sameo, i am ready to delete this function because we will only support MSI.

yangzhon:
@sameo @rbradford,

Please help review this new patches, thanks!

hw/pci-host/pci_lite.c
```diff
@@ -182,7 +183,8 @@ static void pci_lite_realize(DeviceState *dev, Error **errp)
     }
 }
 
-PCIBus *pci_lite_init(MemoryRegion *address_space_mem,
+PCIBus *pci_lite_init(VirtMachineState *vms,
```
sameo: In order to keep this code not `virt` specific, we should pass a `bool nofw` argument to this function instead of bringing a `virt.h` dependency.

hw/pci-host/pci_lite.c
```diff
@@ -209,10 +211,12 @@ PCIBus *pci_lite_init(MemoryRegion *address_space_mem,
...
-    pcie_host_mmcfg_update(pcie, 1, PCI_LITE_PCIEXBAR_BASE,
-                           PCI_LITE_PCIEXBAR_SIZE);
-    e820_add_entry(PCI_LITE_PCIEXBAR_BASE, PCI_LITE_PCIEXBAR_SIZE,
-                   E820_RESERVED);
+    if (!vms->fw) {
+        pcie_host_mmcfg_update(pcie, 1, PCI_LITE_PCIEXBAR_BASE,
+                               PCI_LITE_PCIEXBAR_SIZE);
+        e820_add_entry(PCI_LITE_PCIEXBAR_BASE, PCI_LITE_PCIEXBAR_SIZE,
+                       E820_RESERVED);
+    }
```
sameo: This `pci_lite.c` change is not related to this patch and looks like a fix. It should be moved into its own patch.

sameo (changes requested):
Almost there. I only have 2 comments to address and then we can merge that PR.

hw/pci-host/pci_lite.c
```diff
@@ -215,6 +238,7 @@ static Property pci_lite_props[] = {
                        parent_obj.size, PCI_LITE_PCIEXBAR_SIZE),
     DEFINE_PROP_SIZE(PCI_HOST_PROP_PCI_HOLE64_SIZE, PCILiteHost,
                      pci_hole64_size, DEFAULT_PCI_HOLE64_SIZE),
+    DEFINE_PROP_BOOL("x-pci-hole64-fix", PCILiteHost, pci_hole64_fix, true),
```
rbradford: Why do we need this? You default it to true, and nobody ever sets it to false so please remove the option or remove this.

hw/pci-host/pci_lite.c
yangzhon: okay, thanks, which make code more clean.

hw/pci-host/pci_lite.c
yangzhon: @sameo ,@rbradford

I investigated it again, this e820_add_entry() should be called only in nofw solution and seabios will handle it. We have better verify this with new OVMF bios, let me keep this fix until we do the integrate with new OVMF bios. 

In this time, i will remove this changes.  thanks again!

yangzhon:
@sameo @rbradford,

Please help review this new patches, and the changes as below:
(1). Removed the "x-pci-hole64-fix" option.
(2). Removed e820_add_entry() related changes for PCIe BAR, and keep this fix until when we do the integrate test with new OVMF bios.

thanks!

sameo (approved):

sameo:
LGTM

sameo:
I get the following ACPI tables when running a clear cloud image on top of this PR:

```
[    0.000000] ACPI: RSDP 0x000000001FBFA014 000024 (v02 OVMF  )
[    0.000000] ACPI: XSDT 0x000000001FBF90E8 00003C (v01 OVMF   OVMFEDK2 20130221      01000013)
[    0.000000] ACPI: FACP 0x000000001FBF8000 0000F4 (v03 OVMF   OVMFEDK2 20130221 OVMF 00000099)
[    0.000000] ACPI: DSDT 0x000000001FBF6000 000D57 (v01 INTEL  OVMF     00000004 INTL 20160108)
[    0.000000] ACPI: FACS 0x000000001FBFE000 000040
[    0.000000] ACPI: APIC 0x000000001FBF7000 000090 (v01 OVMF   OVMFEDK2 20130221 OVMF 00000099)
[    0.000000] ACPI: SSDT 0x000000001FBF5000 000057 (v01 REDHAT OVMF     00000001 INTL 20160108)
[    0.000000] ACPI: Local APIC address 0xfee00000
```

We need to fix that before merging.

rbradford (changes requested):
It should look something like:

```
[    0.000000] ACPI: RSDP 0x000000001FBFA014 000024 (v02 BOCHS )
[    0.000000] ACPI: XSDT 0x000000001FBF90E8 000034 (v01 BOCHS  BXPCFACP 00000001      01000013)
[    0.000000] ACPI: FACP 0x000000001FBF7000 00010C (v05 BOCHS  BXPCFACP 00000001 BXPC 00000001)
[    0.000000] ACPI: DSDT 0x000000001FBF8000 00008B (v02 BOCHS  BXPCDSDT 00000001 BXPC 00000001)
[    0.000000] ACPI: APIC 0x000000001FBF6000 000080 (v01 BOCHS  BXPCAPIC 00000001 BXPC 00000001)
```

rbradford:
Here is the OVMF error message: 
```
ProcessCmdAddPointer: invalid pointer location or size in "etc/acpi/tables"
```

yangzhon:
@sameo @rbradford,

This NEW PR fixed the below issue:
ProcessCmdAddPointer: invalid pointer location or size in "etc/acpi/tables"

If we want to debug MMCFG config in guest, please mask below code in hw/pci-host/pci_lite.c
e820_add_entry(PCI_LITE_PCIEXBAR_BASE, PCI_LITE_PCIEXBAR_SIZE,
                   E820_RESERVED);

rbradford (approved):

rbradford:
I've pushed two commits, which with OVMF revision a76b242ca9 will result in a correct memory map in the guest.

rbradford:
@sameo you can test this with https://github.com/rbradford/edk2/releases/download/0.1/OVMF.fd

sameo:
@rbradford works here with your OVMF binary, thanks.
Merging now.

PR #28, opened 2018-07-24
hw: Use dynamic checking casts in callbacks

rbradford:
When casting from void * pointer in callbacks use the casts which
feature a dynamic type check to avoid issues with passing the wrong type
into data pointer for the callback.


sameo (approved):
LGTM

sameo:
It seems our unstable semaphore CI is failing on x86 with this. I'll give it a go tomorrow morning before merging.

sameo:
Manual testing passed.

PR #30, opened 2018-07-26
ACPI CPU hotplug support

sameo:
This is an implementation of the ACPI CPU hotplug for ACPI hardware reduced platforms.
Instead of relying on the SCI interrupt and the GPE registers block, we have to implement support for a Generic Event Device (GED). This is an ACPI hardware reduced platform constraint.

After plumbing the i386/virt machine type with the QEMU CPU hotplug hooks, we have to implement the GED support. We will be using one IRQ per supported GED event type (CPU hotplug first but also memory, PCI and NVDIMM hotplug) and it is up to the virtualized platform to reserve all needed IRQs and provide that as an input to the GED AML building routine. Then when QEMU is being asked to hot-add a CPU, we raise the corresponding IRQ line. The interrupt is trapped by the guest kernel evged driver, which in turn will run the ACPI GED AML code. This will get us back to the same AML code path as the SCI based CPU hotplug, i.e. a complete CPU scan to understand which ones are present and which ones are not.

TODO list:

* [x] Implement i386/virt CPU hotplug hooks
* [x] Implement i386/virt ACPI device hotplug hooks
* [x] Implement ACPI Generic Event Device (GED) support


yangzhon (commented):
static void virt_device_plug_cb(HotplugHandler *hotplug_dev,
                                DeviceState *dev, Error **errp)
{
    VirtAcpiState *s = VIRT_ACPI(hotplug_dev);
     if (object_dynamic_cast(OBJECT(dev), TYPE_CPU)) {
        acpi_cpu_plug_cb(hotplug_dev, &s->cpuhp_state, dev, errp);

The previous acpi_cpu_plug_cb() call gpe event to infom acpi cpu hotplug. In the virt plarform, we need re-implement send_event callback.  below function is still null.
static void virt_send_ged(AcpiDeviceIf *adev, AcpiEventStatusBits ev)
{
}

My question is virt platform does not need inform ACPI_CPU_HOTPLUG_STATUS event to guest kernel in the GED mode?

sameo:
@yangzhon @rbradford @liujing2 This PR is ready now.
With a guest kernel that supports the evged driver, I can hot-add and hot-remove CPUs.

yangzhon:
@sameo,  This PR LGTM!!!

There is only one concern:
(1) GED aml
                /* _CRS interrupt */
                aml_append(crs, aml_interrupt(AML_CONSUMER, AML_LEVEL, AML_ACTIVE_HIGH,
                                              AML_EXCLUSIVE, &events[i].irq, 1));

(2)
static void virt_send_ged(AcpiDeviceIf *adev, AcpiEventStatusBits ev)
{
    VirtAcpiState *s = VIRT_ACPI(adev);
     if (ev & ACPI_CPU_HOTPLUG_STATUS) {
        /* We inject the CPU hotplug interrupt */
        qemu_irq_pulse(s->gsi[VIRT_GED_CPU_HOTPLUG_IRQ]);
    }
}

The GED AML table define the interrupt is level trigger, high level interrupt, but inject interrupt is pulse, which is belong to edge trigger.   if we use qemu_irq_raise() to replace this pulse function? If i am wrong, please correct me, thanks!

hw/i386/virt/virt.c
```diff
@@ -308,74 +308,83 @@ static void virt_2_12_instance_init(Object *obj)
...
 {
     CPUArchId *found_cpu;
     Error *local_err = NULL;
     X86CPU *cpu = X86_CPU(dev);
     MachineState *ms = MACHINE(hotplug_dev);
     VirtMachineState *vms = VIRT_MACHINE(hotplug_dev);
+    HotplugHandlerClass *hhc;
+
+    /* We only support ACPI CPU hotplug/unplug */
+    if (vms->acpi_dev) {
```
rbradford: I know the PC code does this, but in our code is acpi_dev ever NULL? In the PC code ACPI is optional.

rbradford: You use an assert() below...maybe do that instead?

rbradford (approved):
Looks good.

hw/i386/virt/virt.c
sameo: > I know the PC code does this, but in our code is acpi_dev ever NULL? In the PC code ACPI is optional.

Yes, `vms->acpi_dev` can be NULL although it will eventually be equal to `vms->acpi`
 OTOH `vms->acpi` can not be NULL once we boot the guest fw/kernel, so I'll assert that.

sameo:
@rbradford I addressed your comment and we now assert on the `vms->acpi` pointer.

rbradford:
Still lgtm

PR #31, opened 2018-07-30
Enable ARM CI

rbradford:
Make the ARM CI build again and fix build issues that were caused by being unable to test the ARM builds.


rbradford:
@sameo semaphore builds working now. you might want to disable travis as that's the upstream one and it doesn't seem to work right.

sameo (commented):
LGTM

PR #33, opened 2018-08-01
Shutdown and Reboot support

liujing2:
@sameo @rbradford @yangzhon 
I submitted the shutdown/reboot patches to review. Thanks!


hw/acpi/aml-build.c
```diff
@@ -2590,9 +2590,9 @@ void build_fadt(GArray *tbl, BIOSLinker *linker, const AcpiFadtData *f,
     }
 
     /* SLEEP_CONTROL_REG */
-    build_append_gas(tbl, AML_AS_SYSTEM_MEMORY, 0 , 0, 0, 0);
+    build_append_gas_from_struct(tbl, &f->sleep_control_reg);
     /* SLEEP_STATUS_REG */
-    build_append_gas(tbl, AML_AS_SYSTEM_MEMORY, 0 , 0, 0, 0);
+    build_append_gas_from_struct(tbl, &f->sleep_status_reg);
```
rbradford: We don't need this as we are only implementing S5 and the OSPM is never going to read from it. Also you don't initialise this struct so it serves no purpose.

hw/i386/virt/acpi.c
```diff
@@ -81,6 +85,47 @@ static Property virt_acpi_properties[] = {
...
+        }
+    }
+}
+
+static uint64_t virt_acpi_sleep_cnt_read(void *opaque, hwaddr addr,
+                                         unsigned width)
+{
+    VirtAcpiState *d = opaque;
+
+    return d->sleep_control_regs;
```
rbradford: The ACPI specification never talks about reading from this register. Have you seen code that expects to read from this register? If not please don't implement read on the ioport (which means you can remove the sleep_control_regs variable.)

include/hw/acpi/reduced.h
```diff
@@ -1,6 +1,9 @@
 #ifndef HW_ACPI_REDUCED_H
 #define HW_ACPI_REDUCED_H
 
+#define ACPI_REDUCED_SLP_CNT_IOPORT 0x3B0
+#define ACPI_REDUCED_SLEEP_ENABLE 0x0020 /* SLP_EN */
```
rbradford: ACPI_REDUCED_SLEEP_ENABLE is only used in acpi.c so can be in that .c file i would also prefer (1 << 5) to 0x20

include/hw/acpi/reduced.h
```diff
@@ -1,6 +1,9 @@
 #ifndef HW_ACPI_REDUCED_H
 #define HW_ACPI_REDUCED_H
 
+#define ACPI_REDUCED_SLP_CNT_IOPORT 0x3B0
```
rbradford: How did you choose this ioport? Also as the field is called SLEEP_CONTROL can you use the full name rather than shortening.

hw/i386/virt/acpi.c
```diff
@@ -81,6 +85,47 @@ static Property virt_acpi_properties[] = {
...
+    .read = virt_acpi_sleep_cnt_read,
+};
+
+static void virt_acpi_realizefn(DeviceState *dev, Error **errp)
+{
+    SysBusDevice *sys = SYS_BUS_DEVICE(dev);
+    VirtAcpiState *d = VIRT_ACPI(dev);
+
+    memory_region_init_io(&d->sleep_iomem, OBJECT(dev),
+                          &virt_sleep_cnt_ops, d, TYPE_VIRT_ACPI, 2);
```
rbradford: You reserve two io-ports here. You should pass 1 as the last parameter.

hw/i386/virt/acpi.c
```diff
@@ -124,6 +153,10 @@ static void virt_acpi_realizefn(DeviceState *dev, Error **errp)
     memory_region_init_io(&d->sleep_iomem, OBJECT(dev),
                           &virt_sleep_cnt_ops, d, TYPE_VIRT_ACPI, 2);
     sysbus_add_io(sys, ACPI_REDUCED_SLP_CNT_IOPORT, &d->sleep_iomem);
+
+    memory_region_init_io(&d->reset_iomem, OBJECT(dev),
+                          &virt_reset_cnt_ops, d, TYPE_VIRT_ACPI, 2);
```
rbradford: Again, two i/o ports rather than one.

include/hw/acpi/reduced.h
```diff
@@ -2,6 +2,7 @@
 #define HW_ACPI_REDUCED_H
 
 #define ACPI_REDUCED_SLP_CNT_IOPORT 0x3B0
+#define ACPI_REDUCED_RST_CNT_IOPORT 0x3C0
```
rbradford: The register is called RESET_REG in the spec (no CONTROL.)

hw/i386/virt/acpi.c
```diff
@@ -111,11 +113,38 @@ static uint64_t virt_acpi_sleep_cnt_read(void *opaque, hwaddr addr,
...
+{
+    VirtAcpiState *d = opaque;
+
+    return d->reset_control_regs;
+
+}
+
+static const MemoryRegionOps virt_reset_cnt_ops = {
+    .write = virt_acpi_reset_cnt_write,
+    .read = virt_acpi_reset_cnt_read,
```
rbradford: Is read necessary here? I don't think so.

hw/i386/virt/acpi.c
```diff
@@ -111,11 +113,38 @@ static uint64_t virt_acpi_sleep_cnt_read(void *opaque, hwaddr addr,
...
     .write = virt_acpi_sleep_cnt_write,
     .read = virt_acpi_sleep_cnt_read,
 };
 
+static void virt_acpi_reset_cnt_write (void *opaque, hwaddr addr,
+                                      uint64_t val, unsigned width)
+{
+    VirtAcpiState *d = opaque;
+
+    if (val & 4) {
```
rbradford: This needs to match the value in RESET_VALUE from the FADT, i don't see you setting that so you must be reusing it. Can you add a #define to make it easier to understand (and use that in the FADT table.)

Also RESET_REG_SUP in the FADT needs to be enabled..is it?

hw/i386/virt/acpi.c
```diff
@@ -81,6 +85,47 @@ static Property virt_acpi_properties[] = {
...
+                                      uint64_t val, unsigned width)
+{
+    VirtAcpiState *d = opaque;
+    uint16_t sus_type = (val >> 2) & 7;
+
+    d->sleep_control_regs = val & ~(ACPI_REDUCED_SLEEP_ENABLE);
+
+    if (val & ACPI_REDUCED_SLEEP_ENABLE) {
+        switch (sus_type) {
+        case 5:
```
rbradford: the \_S5 object in the DSDT needs to be set to the same value othewise this won't work for the Linux kernel.

rbradford (changes requested):
Looking good. Some small changes please.

hw/acpi/aml-build.c
liujing2: OK, got it. Will remove it.

hw/i386/virt/acpi.c
liujing2: When OVMF writes 2:5 bits (using function IoBitFieldWrite16(), IoOr16 ), it need firstly read from the ioport and then write.

include/hw/acpi/reduced.h
liujing2: This port is originally used by "0x03B0-0x03DF : The range used for the IBM VGA".  I think virt platform won't use this device. 
OK, I will change the name to be full.

hw/i386/virt/acpi.c
liujing2: Will change to 1.

hw/i386/virt/acpi.c
liujing2: Yes, the same. Will change.

include/hw/acpi/reduced.h
liujing2: Oh, got it.

hw/i386/virt/acpi.c
liujing2: Hmm, according to the ovmf changes, I think this read callback is useless. I am not sure maybe we can discuss the ovmf changes firstly. Qemu depends on it.

liujing2: And I read the ich9_rst_cnt_read/ich9_rst_cnt_write in hw/isa/lpc-ich9.c, it indicates something which I didn't find out in the ACPI spec, like FULL_RST (bit 3) and SYS_RST (bit1). Also piix shows similarly(rcr_read/rcr_write).  And I gdb pc reboot and found rcr_read not used.
Do you have some ideas on that?

hw/i386/virt/acpi.c
liujing2: Yep, RESET_REG_SUP will be set. Will do it.
Also, I'd like to discuss the RESET_VALUE definition. q35 defines it as 0xf and use 4 to set guest reset. I followed it but this confused me a lot... Of course, we could #define a value, set it to fadt.reset_val and use the #define in write callback.

hw/i386/virt/acpi.c
liujing2: OK, will add the _S5 object.

hw/i386/virt/acpi.c
rbradford: Why does it need to read? The specification at no point says the bits need to be unchanged.

hw/i386/virt/acpi.c
rbradford: For RESET_REG you can choose any value provided you advertise it correctly in the FADT.

hw/i386/virt/acpi.c
liujing2: The same to you. I didn't find any spec about the read operations. hmm, this could be done by changing OVMF implementation as follows. But in this way, OVMF can not show clearly what the bits are for. 

```
--- a/OvmfPkg/Library/ResetSystemLib/ResetSystemLib.c
+++ b/OvmfPkg/Library/ResetSystemLib/ResetSystemLib.c
@@ -62,8 +62,7 @@ AcpiSleepControl (

   AcpiReducedPort = 0x3B0;

-  IoBitFieldWrite16 (AcpiReducedPort, 2, 5, (UINT16) SuspendType);
-  IoOr16 (AcpiReducedPort, BIT5);
+  IoWrite16(AcpiReducedPort, BIT2 | BIT4 | BIT5);
   CpuDeadLoop ();
```

include/hw/acpi/reduced.h
liujing2: Will change that in the latest codes in PR.

hw/i386/virt/acpi.c
liujing2: Will change the value to a #define

liujing2:
@rbradford I have updated all the discussion changes. Thanks!

hw/i386/virt/acpi.c
rbradford: @liujing2 I don't understand your complaint. Reading the value has no benefit. For sleeping you're supposed to write the value from \_S5 into the registere ORed with the enable bit. Because you're not reading that value i'm happy for you to just write SLEEP_LEVEL << 2 | SLEEP_EN into the register. Where in OVMF you #define SLEEP_LEVEL to 5.

liujing2: OK, updated.
I am finding how to get FADT in OVMF. By gEfiAcpiTableGuid (defined in ACPI spec 5.2.5.2) and EfiGetSystemConfigurationTable(), probably we could get the RSDP firstly... I haven't investigated if we could reuse the function.

rbradford (approved):
Looks good. How did you test this?

rbradford:
I've tested this with the OVMF changes. I want some minor changes to those (constants) but this code is lgtm so can be merged.

liujing2:
@rbradford  yep, I tested it using OVMF changes.

PR #36, opened 2018-08-03
Topic/virt x86: memory hotplug support with GED

yangzhon:
@sameo @mcastelino @rbradford @liujing2 

Like CPU acpi hotplug did, we add the memory hotplug with GED solution.  This PR is in WIP, and please help review. thanks!

PR #37, opened 2018-08-03
acpi: reduced: Separate GED device creation AML from the CPUs one

sameo:
We are currently calling the GED AML build from the CPUs addition one.
They're not directlt related, so we should just separate them out.


yangzhon (approved):
LGTM.

My PR #36 also did almost same thing.

PR #39, opened 2018-08-03
Virt hotplug fixes

sameo:
A couple of fixes for the virt hotplug implementation:

- Do not error out on unsupported devices pre-plug hooks, as not all devices require a pre-plug hook implementation
- Do not create the legacy AML hotplug event method when running HW reduced ACPI.


yangzhon (approved):
LGTM!

Thanks!

yangzhon (approved):
LGTM!

This setting can fix [GED_] namespace error in guest kernel, thanks!

Issue #16, opened 2018-06-29
virtio support for i386/virt


sameo:
We want the i386 virt platform to support virtio devices.
As we want to keep the PCI dependency to a minimum, we should aim at exposing virtio devices to the guest through MMIO.

TBD: If we are to support minimal PCI emulation (host bridge+root port), can we go virtio pci instead?

liujing2:
@rbradford @sameo @mcastelino @yangzhon 
I tested virtio-blk-pci and virtio-net-pci hotplugging onto **pcie-root-port** on i386/virt platform and succeed.
Is this what we expected?

Commands:
{ "execute": "device_add", "arguments": { "driver": "virtio-net-pci", "id": "virtio-net1", "bus": "rp1", "bootindex": "2", "addr": "2" } }

{ "execute": "blockdev-add", "arguments":{"driver":"raw","node-name":"disk1", "file":{"driver":"file","filename":"/home/liujing/work/image/null.img" } } }
{ "execute": "device_add", "arguments": { "driver": "virtio-blk", "drive": "disk1", "id": "virtio1", "bus": "rp1", "bootindex": "1", "addr": "1"} }

rbradford:
/CC @sboeuf

I thought Sebastien tried this yesterday and found it didn't work

sboeuf:
@liujing2 what is the rp1 bus ? Is that something that has been introduced with some recent OVMF changes ?

What I did yesterday was:
```bash
sudo ./x86_64-softmmu/qemu-system-x86_64 \
    -bios $HOME/workloads/OVMF.fd \
    -nographic \
    -nodefaults \
    -L . \
    -machine virt,accel=kvm,kernel_irqchip \
    -smp sockets=1,cpus=4,cores=2,maxcpus=8 -cpu host \
    -m 8192M,slots=2,maxmem=16950M \
    -device sysbus-debugcon,iobase=0x402,chardev=debugcon -chardev file,path=/tmp/debug-log,id=debugcon \
    -device virtio-blk-pci,drive=image -drive if=none,id=image,file=$HOME/workloads/clear-24250-cloud.img \
    -device virtio-serial-pci,id=virtio-serial0 -device virtconsole,chardev=charconsole0,id=console0 -chardev stdio,id=charconsole0 \
    -netdev user,id=mynet0,hostfwd=tcp::${VMN}0022-:22,hostfwd=tcp::${VMN}2375-:2375 -device virtio-net-pci,netdev=mynet0 \
[...]

PR #41, opened 2018-08-08
hw: i386: virt: Add NVDIMM support

### LINKED ISSUE #44: NVDIMM support
We need to implement the NVDIMM support for cases that would require the image being passed as a fake NVDIMM.

sboeuf:
This commit introduces the NVDIMM support. It includes the NVDIMM
hotplug since the way NVDIMM is added is always through hotplug,
even when an NVDIMM device is defined from the Qemu command line.

First, this code builds a new SSDT table with the NVDR device
described inside as a specific platform device.

Then it updates the GED method of the DSDT table, by defining a new
Notify() method to notify the guest OS driver that something changed
regarding the NVDIMM devices.

All of this happen if NVDIMM is enabled at the machine level.

The last thing this patch does is to send an interrupt to the guest
kernel everytime a new NVDIMM device is plugged. This triggers the
GED method defined in the DSDT.

Fixes #44

### DIFF
--- a/hw/acpi/reduced.c
+++ b/hw/acpi/reduced.c
@@ -187,6 +187,11 @@ static void acpi_reduced_build(MachineState *ms, AcpiBuildTables *tables, AcpiCo
         acpi_add_table(table_offsets, tables_blob);
         mc->firmware_build_methods.acpi.mcfg(tables_blob, tables->linker, &mcfg);
     }
+    if (conf->acpi_nvdimm_state.is_enabled) {
+        nvdimm_build_acpi(table_offsets, tables_blob, tables->linker,
+                          &conf->acpi_nvdimm_state, ms->ram_slots);
+    }
+
     /* RSDT is pointed to by RSDP */
     xsdt = tables_blob->len;
     build_xsdt(tables_blob, tables->linker, table_offsets, NULL, NULL);
@@ -326,8 +331,7 @@ static Aml *ged_event_aml(GedEvent *event)
         return aml_call0("\\_SB.MHPC." MEMORY_SLOT_SCAN_METHOD);
     case GED_PCI_HOTPLUG:
     case GED_NVDIMM_HOTPLUG:
-        /* Not supported for now */
-        return NULL;
+        return aml_notify(aml_name("\\_SB.NVDR"), aml_int(0x80));
     default:
         break;
     }
--- a/hw/i386/virt/acpi.c
+++ b/hw/i386/virt/acpi.c
@@ -65,8 +65,12 @@ static void virt_device_plug_cb(HotplugHandler *hotplug_dev,
     if (object_dynamic_cast(OBJECT(dev), TYPE_CPU)) {
         acpi_cpu_plug_cb(hotplug_dev, &s->cpuhp_state, dev, errp);
     } else if (object_dynamic_cast(OBJECT(dev), TYPE_PC_DIMM)) {
-        acpi_memory_plug_cb(hotplug_dev, &s->memhp_state,
-                            dev, errp);
+        if (object_dynamic_cast(OBJECT(dev), TYPE_NVDIMM)) {
+            nvdimm_acpi_plug_cb(hotplug_dev, dev);
+        } else {
+            acpi_memory_plug_cb(hotplug_dev, &s->memhp_state,
+                                dev, errp);
+        }
     }  else {
         error_setg(errp, "virt: device plug request for unsupported device"
                    " type: %s", object_get_typename(OBJECT(dev)));
@@ -117,6 +121,8 @@ static void virt_send_ged(AcpiDeviceIf *adev, AcpiEventStatusBits ev)
     } else if (ev & ACPI_MEMORY_HOTPLUG_STATUS) {
         /* We inject the memory hotplug interrupt */
         qemu_irq_pulse(s->gsi[VIRT_GED_MEMORY_HOTPLUG_IRQ]);
+    } else if (ev & ACPI_NVDIMM_HOTPLUG_STATUS) {
+        qemu_irq_pulse(s->gsi[VIRT_GED_NVDIMM_HOTPLUG_IRQ]);
     }
 }
 
--- a/hw/i386/virt/virt.c
+++ b/hw/i386/virt/virt.c
@@ -100,7 +100,8 @@ static void acpi_conf_virt_init(MachineState *machine, AcpiConfiguration *conf)
     conf->acpi_dev = vms->acpi_dev;
     conf->cpu_hotplug_io_base = VIRT_CPU_HOTPLUG_IO_BASE;
     conf->hotplug_memory = vms->hotplug_memory;
-
+    conf->acpi_nvdimm_state = vms->acpi_nvdimm_state;
+ 
     /* GED events */
     GedEvent events[] = {
         {
@@ -110,6 +111,10 @@ static void acpi_conf_virt_init(MachineState *machine, AcpiConfiguration *conf)
         {
             .irq   = VIRT_GED_MEMORY_HOTPLUG_IRQ,
             .event = GED_MEMORY_HOTPLUG,
+        },
+        {
+            .irq   = VIRT_GED_NVDIMM_HOTPLUG_IRQ,
+            .event = GED_NVDIMM_HOTPLUG,
         },
     };
 
@@ -228,6 +233,11 @@ static void virt_machine_state_init(MachineState *machine)
                             val, sizeof(*val));
         }
 
+        if (vms->acpi_nvdimm_state.is_enabled) {
+            nvdimm_init_acpi_state(&vms->acpi_nvdimm_state, get_system_io(),
+                                   fw_cfg, OBJECT(vms));
+        }
+
         vms->fw_cfg = fw_cfg;
         acpi_conf_virt_init(MACHINE(vms), vms->acpi_configuration);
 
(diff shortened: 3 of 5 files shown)


sboeuf:
Still work in progress.
@rbradford @mcastelino @yangzhon PTAL.
I run into some weird segfault:
```
0x000055555598fa5a in bios_linker_loader_add_pointer (linker=linker@entry=0x555556c4cea0, dest_file=dest_file@entry=0x555555cf8791 "etc/acpi/tables", dst_patched_offset=dst_patched_offset@entry=4845, dst_patched_size=dst_patched_size@entry=4 '\004', 
    src_file=src_file@entry=0x555555d1986d "etc/acpi/nvdimm-mem", src_offset=src_offset@entry=0) at hw/acpi/bios-linker-loader.c:191
191         assert(src_offset < source_file->blob->len);
(gdb) bt
#0  0x000055555598fa5a in bios_linker_loader_add_pointer (linker=linker@entry=0x555556c4cea0, dest_file=dest_file@entry=0x555555cf8791 "etc/acpi/tables", dst_patched_offset=dst_patched_offset@entry=4845, dst_patched_size=dst_patched_size@entry=4 '\004', 
    src_file=src_file@entry=0x555555d1986d "etc/acpi/nvdimm-mem", src_offset=src_offset@entry=0) at hw/acpi/bios-linker-loader.c:191
#1  0x000055555598e18f in nvdimm_build_ssdt (ram_slots=3, dsm_dma_arrea=0x0, linker=0x555556c4cea0, table_data=0x555556a1bca0, table_offsets=0x555556a1bd90) at hw/acpi/nvdimm.c:1280
#2  nvdimm_build_acpi (table_offsets=table_offsets@entry=0x555556a1bd90, table_data=table_data@entry=0x555556a1bca0, linker=0x555556c4cea0, state=state@entry=0x555556c4c120, ram_slots=3) at hw/acpi/nvdimm.c:1300
#3  0x000055555598f210 in acpi_reduced_build (ms=ms@entry=0x555556a1c5d0, tables=tables@entry=0x7fffffffdfe0, conf=conf@entry=0x555556c4c0e0) at hw/acpi/reduced.c:187
#4  0x000055555598f314 in acpi_reduced_setup (machine=0x555556a1c5d0, conf=0x555556c4c0e0) at hw/acpi/reduced.c:280
#5  0x0000555555bfcb64 in notifier_list_notify (list=<optimized out>, data=0x0) at util/notify.c:40
#6  0x00005555557f0adb in qemu_run_machine_init_done_notifiers () at vl.c:2723
#7  main (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at vl.c:4689
``` 
Let me know if you understand what's going wrong here. Basically, the function `bios_linker_find_file()` return NULL, which causes the segfault here, but I don't understand why the entry cannot be found.

yangzhon:
@sboeuf ,

How did you test this ? with nofw or OVMF binary ?  You can test it firstly on nofw solution and then verify it with OVMF.

As for you issue, it seems that  source_file->blob->len=0, need to check this length. thanks!

sboeuf:
@yangzhon I've tested with both, and in both cases I end up with this segfault. What's the reason the blob length would be 0?

sboeuf:
@rbradford greatly helped me debugging this segmentation fault issue and we found that we were not providing an uninitialized pointer at some point. This was due to the confusion of copying everything between `vms` and `vms->acpi_configuration`.

sboeuf:
I have tested the PR with @sameo kernel config [here](https://gist.githubusercontent.com/sameo/657d5430cd11fe8a05266a28b83568b0/raw/809d7f85469ddbbfd691e9aa9a33b56a5e5507b4/cc-nemu-kernel-config.txt). The most important being to have the NVDIMM configs enabled (`CONFIG_LIBNVDIMM=y`, `CONFIG_BLK_DEV_PMEM=y`, `CONFIG_ND_BLK=y`, `CONFIG_ACPI_NFIT=y`):
```bash
sudo ./x86_64-softmmu/qemu-system-x86_64    -nographic      -nodefaults      -L .      -net none      -machine virt,accel=kvm,kernel_irqchip,nvdimm,nofw      -smp sockets=1,cpus=4,cores=2,maxcpus=8 -cpu host      -m 2G,slots=3,maxmem=16G      -kernel /home/sebastien/workloads/linux/arch/x86/boot/compressed/vmlinux.bin -append 'console=hvc0 single iommu=false root=/dev/ram0'      -device virtio-serial-pci,id=virtio-serial0 -device virtconsole,chardev=charconsole0,id=console0 -chardev stdio,id=charconsole0      -monitor telnet:127.0.0.1:55555,server,nowait
```
and I have the SSDT tables as expected.
Also, after I have hotplugged 3 NVDIMM devices:
```
QEMU 2.12.0 monitor - type 'help' for more information
(qemu) object_add memory-backend-file,id=mem4,share=on,mem-path=/home/sebastien/workloads/devmap.img,size=52428800
(qemu) device_add nvdimm,id=nvdimm4,memdev=mem4
...
```
I have the expected result in `/proc/iomem` where I can find the 3 `Persistent Memory` address spaces:
```bash
/ # cat /proc/iomem 
00000000-00000fff : reserved
00001000-0009ffff : System RAM
000a0000-000bffff : PCI Bus 0000:00
000e0000-000fffff : reserved
  000f0000-000fffff : System ROM
00100000-7ffbffff : System RAM
  01000000-015c8b02 : Kernel code
  015c8b03-0189b9bf : Kernel data
  01aca000-01b67fff : Kernel bss
7ffc0000-8fffffff : reserved
  80000000-8fffffff : PCI MMCONFIG 0000 [bus 00-ff]
90000000-fec00000 : PCI Bus 0000:00
  90000000-90000fff : 0000:00:01.0
fee00000-fee00fff : Local APIC
feffc000-feffffff : reserved
100000000-1031fffff : Persistent Memory
  100000000-1031fffff : namespace0.0
103200000-1063fffff : Persistent Memory
  103200000-1063fffff : namespace1.0
106400000-1095fffff : Persistent Memory
  106400000-1095fffff : namespace2.0
540000000-d3fffffff : PCI Bus 0000:00
  540000000-540003fff : 0000:00:01.0
    540000000-540003fff : virtio-pci-modern
```

@yangzhon @rbradford @mcastelino PTAL and let me know if something's wrong !

sboeuf:
A few notes here:
With this PR, I have been able to run Kata containers from Docker. First the `configuration.toml` needs to be slightly modified:
- qemu `path` has to be updated to the qemu binary generated from nemu repo.
- `kernel` has to be `vmlinux` instead of `vmlinuz`
- `machine_type` has to be `virt` instead of `pc`
- `machine_accelerators` has to be `nofw` instead of being empty

Then, we need this [PR](https://github.com/kata-containers/runtime/pull/559) to be merged on the runtime repo.

__Use cases__
First, I want to mention it works both with `vsock` and `virtserialport`, which are the two different modes `without` and `with` the proxy respectively.

Now, as long as we use `virtio-scsi` for our block devices (the default for Kata), the hotplug works because it is directly handled by the scsi controller. Instead, if we use `virtio-blk`, the hotplug won't work as we get the following error:
```
msg="{\"error\": {\"class\": \"GenericError\", \"desc\": \"Bus 'pcie.0' does not support hotplugging\"}}"
```
This also means that because we don't support PCI hotplug yet, `vfio-pci` won't work.

One thing about the storage drivers, it works both with `overlay` (ending up using virtio-9p), and `devicemapper` (ending up using virtio-scsi).

/cc @sameo @mcastelino @rbradford @yangzhon

yangzhon:
@sboeuf 

The virt can be merged with Kata container, great!!!

As for "'pcie.0' does not support hotplugging", please use native pcie hotplug, we will only support this kind of hotplug. Please add below command in your qemu command:
-device pcie-root-port,id=root1

#16  , @liujing2 has updated the virtio hotplug results.

rbradford (approved):
lgtm from me. Can you look into the semaphore virt failure and then I can merge?

sboeuf:
More debug with @rbradford today. The fix that we applied yesterday regarding the segmentation fault was the culprit of the CI failure because it was using a pointer that was initialized a bit later. It was only failing for the OVMF case because the ACPI tables were corrupted and replaced by the firmware itself.

Moving the code right in between solved the issue with a little duplication that might be refactored later as part of a PR doing some more general refactoring.

sboeuf:
One last note, the Clear Linux cloud image works too. We need the appropriate kernel config to get it supported. Here is the patch I've applied to the config:
```diff
+diff --git a/config b/config
index 53adbed..0168750 100644
--- a/config
+++ b/config
@@ -669,8 +669,8 @@ CONFIG_ACPI_HOTPLUG_IOAPIC=y
 # CONFIG_ACPI_HED is not set
 # CONFIG_ACPI_CUSTOM_METHOD is not set
 # CONFIG_ACPI_BGRT is not set
-# CONFIG_ACPI_REDUCED_HARDWARE_ONLY is not set
-# CONFIG_ACPI_NFIT is not set
+CONFIG_ACPI_REDUCED_HARDWARE_ONLY=y
+CONFIG_ACPI_NFIT=y
 CONFIG_HAVE_ACPI_APEI=y
 CONFIG_HAVE_ACPI_APEI_NMI=y
 # CONFIG_ACPI_APEI is not set
@@ -3005,19 +3005,19 @@ CONFIG_ARM_GIC_MAX_NR=1
 # Android
 #
 # CONFIG_ANDROID is not set
-CONFIG_LIBNVDIMM=m
-CONFIG_BLK_DEV_PMEM=m
-CONFIG_ND_BLK=m
+CONFIG_LIBNVDIMM=y
+CONFIG_BLK_DEV_PMEM=y
+CONFIG_ND_BLK=y
 CONFIG_ND_CLAIM=y
-CONFIG_ND_BTT=m
+CONFIG_ND_BTT=y
 CONFIG_BTT=y
-CONFIG_ND_PFN=m
+CONFIG_ND_PFN=y
 CONFIG_NVDIMM_PFN=y
 CONFIG_NVDIMM_DAX=y
 CONFIG_DAX_DRIVER=y
 CONFIG_DAX=y
 CONFIG_DEV_DAX=y
-CONFIG_DEV_DAX_PMEM=m
+CONFIG_DEV_DAX_PMEM=y
 # CONFIG_NVMEM is not set
```

yangzhon:
This PR is LGTM.

Once this PR is merged, i will rebase numa patchset, thanks!

PR #43, opened 2018-08-08
[BUG]: Fixed two bugs.

yangzhon:
@sameo @rbradford @mcastelino

I fixed two bugs, one is related with cpu aml table and another is loading bzImage coredump issue in pc/q35 platform. please help review those two patches, thanks!


rbradford:
@yangzhon have you tested with Q35 non-direct boot. It looks like it fails from the semaphore logs.

hw/acpi/reduced.c
```diff
@@ -60,15 +60,6 @@ static void acpi_dsdt_add_memory_hotplug(MachineState *ms, Aml *dsdt)
 
 static void acpi_dsdt_add_cpus(MachineState *ms, Aml *dsdt, Aml *scope, int smp_cpus, AcpiConfiguration *conf)
 {
-    uint16_t i;
-
-    for (i = 0; i < smp_cpus; i++) {
-        Aml *dev = aml_device("C%.03X", i);
-        aml_append(dev, aml_name_decl("_HID", aml_string("ACPI0007")));
```
sboeuf: Just curious, do you know why this was added in the first place ?

yangzhon: @sboeuf ,

This was added by samuel and the subsequent build_cpus_aml() function also do this.  which made below error log in guest
acpi ACPI0007:08: BIOS reported wrong ACPI id 0 for the processor

yangzhon:
@sameo @rbradford @mcastelino

For patch "bzImage: fixed load bzImage coredump in q35/pc", i suggested we hold on this patch.

I debugged this issue and found all bzImage can not work in xen/pc/q35/virt-i386 platform.  Since the acpi_conf_pc_init() is in pc_machine_done(), when the load_linux_bzimage() is called by xen/pc/q35, the conf struct is not initialized. 

In fact, the loading bzImage in virt-i386 also have some issues, although virt platform has called acpi_conf_virt_init(), but conf->acpi_data_size and conf->linuxboot_dma_enabled are never initialized in virt.c file.

So, my question is below:
If virt-i386 need support bzImage? If yes, we had better re-design load_linux_bzimage()'s parameters.
If no, we can remove load_linux_bzimage() in virt-i386 and restore the load_linux_bzimage()'s previous parameters.

rbradford:
I think this PR is not showing the correct code now as the branch has changed.

PR #56, opened 2018-08-20
WIP - NATS

rbradford:
Natty by name, natty by nature. Here is what i've got so far for some testing.

TODO:

- [ ] Need to fetch the right files into the right locations (for the image, OVMF, etc)
- [ ] Setup the QMP socket and connect the qmp code from govmm (need to do that in parallel to the goroutine running qemu so might need some refactoring)
- [ ] Audit the qemu args
- [ ] Add some tests around hotplugging
- [ ] ....?


rbradford:
Not actually expecting a review at this point, more just an FYI.

sameo (approved):
LGTM

PR #58, opened 2018-08-22
hw: i386: virt: Add support for PCI hotplug through ACPI

sboeuf:
This PR enables the support for hotplugging PCI devices through ACPI:
- Enable PCI bus to accept hotplugged devices, by providing an
appropriate hotplug handler.
- Implement the code path generating the appropriate interrupt in
case a new PCI device is getting added to the PCI bus.
- Modify the definition of the PCI bus in the DSDT table to make
sure a PCI scan of all the available slots can be performed when
the interrupts comes in.
- Modify GED device in the DSDT table, so that an interrupt related
to PCI will trigger a new PCI scan.

Part of this work include some refactoring and the removal of unneeded
code related to the pc machine type.

Signed-off-by: Jing Liu <jing2.liu@linux.intel.com>


sboeuf:
@liujing2 here is some work in progress on my side about PCI devices.
Just to summarize this, I can hotplug some pci devices to `pcie.0`, but I need to rescan manually the PCI bus from inside the guest to make the device appear.
I am working on the GED support, and how to make the ACPI trigger the rescan when it receives the interrupt.

liujing2:
@sboeuf  wow, your codes almost done the key parts of acpi pci hotplug. I also prepared one #59 but have some differences and met problems now. And I prefer yours. 
I looked into the codes and got why we have to manually rescan. I wrote a patch for it and submitted PR #60. Now acpi pci hotplug onto pcie.0 can be done successfully. 
BTW, I think we need some function factoring.  I will take a look later today.

sboeuf:
@liujing2 I have completed the rework of this PR, including some factorization and some cleanup. I think we can close #59 and #60 as this PR supersedes them. WDYT?

@sameo @mcastelino @rbradford @yangzhon @liujing2 PTAL

Issue #65, opened 2018-08-28
3.0.0 breaks pci windows


liujing2:
@sameo @mcastelino @rbradford @sboeuf @yangzhon 
dmesg  shows cold-plugged devices are assigned by firmware.
And if we add a pcie-root-port when launch guest, it also failed for "no space"
```

[    0.021357] pnp: PnP ACPI init
[    0.021386] pnp: PnP ACPI: found 0 devices
[    0.022620] pci_bus 0000:00: max bus depth: 0 pci_try_num: 1
[    0.022623] pci 0000:00:01.0: BAR 4: no space for [mem size 0x00004000 64bit pref]
[    0.022625] pci 0000:00:01.0: BAR 4: trying firmware assignment [mem 0x800000000-0x800003fff 64bit pref]
[    0.022627] pci 0000:00:01.0: BAR 4: assigned [mem 0x800000000-0x800003fff 64bit pref]
[    0.023374] pci 0000:00:02.0: BAR 4: no space for [mem size 0x00004000 64bit pref]
[    0.023375] pci 0000:00:02.0: BAR 4: trying firmware assignment [mem 0x800004000-0x800007fff 64bit pref]
[    0.023377] pci 0000:00:02.0: BAR 4: assigned [mem 0x800004000-0x800007fff 64bit pref]
[    0.024108] pci 0000:00:01.0: BAR 1: no space for [mem size 0x00001000]
[    0.024109] pci 0000:00:01.0: BAR 1: trying firmware assignment [mem 0x90001000-0x90001fff]
[    0.024110] pci 0000:00:01.0: BAR 1: assigned [mem 0x90001000-0x90001fff]
[    0.024122] pci 0000:00:02.0: BAR 1: no space for [mem size 0x00001000]
[    0.024123] pci 0000:00:02.0: BAR 1: trying firmware assignment [mem 0x90000000-0x90000fff]
[    0.024124] pci 0000:00:02.0: BAR 1: assigned [mem 0x90000000-0x90000fff]
[    0.024133] pci 0000:00:01.0: BAR 0: no space for [io  size 0x0040]
[    0.024134] pci 0000:00:01.0: BAR 0: trying firmware assignment [io  0xc040-0xc07f]
[    0.024135] pci 0000:00:01.0: BAR 0: assigned [io  0xc040-0xc07f]
[    0.024145] pci 0000:00:02.0: BAR 0: no space for [io  size 0x0040]
[...]

PR #67, opened 2018-08-28
hw/i386/virt: Initialize AcpiConfiguration's below_4g_mem_size

rbradford:
This struct member is read by the ACPI table (SRAT) population code in order to
publish the PCI hole in the NUMA configuration.


rbradford:
I found this when looking at #65 after looking at the diff between the two branches.

mcastelino (approved):
LGTM

PR #66, opened 2018-08-28
hw/acpi: Fix PCI host bridge ACPI initialisation

rbradford:
The function build_pci_host_bridge() formerly added the newly created scope to
the table however as it now returns the scope instead it is the caller's
responsibility to add to the table.

This fixes the error message on bootup:

[    0.007492] acpi PNP0A08:00: [Firmware Bug]: no secondary bus range in _CRS

And prevented the BARs from being allocated for the PCI devices on the bus

Signed-off-by: Rob Bradford <robert.bradford@intel.com>

PR #68, opened 2018-08-29
dsdt: simplify PCI0 path

liujing2:
@mcastelino @sameo  @rbradford @sboeuf
 
The first patch is based on #66 
The second patch simplifies the PCI0 device path from \_SB._SB.PCI0 to \_SB.PCI0


rbradford (approved):
I looked at the iasl output and it looks better.

PR #73, opened 2018-09-03
nats: Add CPU and memory hotplug support

rbradford:
Test basic CPU and memory hotplug with NATS.


rbradford:
Might split this PR as the cpu hotplug test seems to fail on Jenkins (but runs fine locally....)

PR #74, opened 2018-09-03
nats: Add a test for memory hotplug

rbradford:
Test that memory hotplug works by seeing if the MemTotal in the guest
increases after inserting a DIMM.

### DIFF
--- a/tools/CI/nats/nemu_test.go
+++ b/tools/CI/nats/nemu_test.go
@@ -10,6 +10,7 @@ import (
 	"os/user"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


rbradford:
Memory only for now..CPU hotplug is consistently failing on the Jenkins VM so need to root-cause that separatelly.

tools/CI/nats/nemu_test.go
```diff
@@ -398,3 +399,49 @@ func TestQMPQuit(t *testing.T) {
...
+		ctx, cancelFunc := context.WithTimeout(context.Background(), 120*time.Second)
+		err := q.startQemu(ctx, t)
+		if err != nil {
+			cancelFunc()
+			<-q.doneCh
+			t.Fatalf("Error starting qemu: %v", err)
+		}
+
+		beforeMem := getTotalMemory(t)
+		err = q.qmp.ExecHotplugMemory(ctx, "memory-backend-ram", "memslot1", "", 512)
```
sameo: One nit: Could we define those `512` and `1024` as constants? We may need to reuse them for hot unplug tests...

rbradford:
@sameo updated.

sameo (approved):
LGTM

PR #77, opened 2018-09-05
tools: Run NATS locally

sameo:
With a ccloudvm VM, we can setup and build NEMU, and then run NATS
locally. This calls run_nats.sh under the same conditions as our Jenkins based CI.

### DIFF
--- a/tools/CI/run_nats.sh
+++ b/tools/CI/run_nats.sh
@@ -1,38 +1,50 @@
 #!/bin/bash
-set -e
 set -x
+
+GO_VERSION="1.10.3"
+CLEAR_VERSION=24740
+CLEAR_IMAGE=clear-$CLEAR_VERSION-cloud.img.xz
+WORKLOADS_DIR="$HOME/workloads"
+OVMF="OVMF.fd"
+
+go_install() {
+    export PATH=/usr/local/go/bin:$PATH
+    go version | grep $GO_VERSION
+    if [ $? -ne 0 ]; then
+	pushd /tmp
+	wget -nv https://dl.google.com/go/go1.10.3.linux-amd64.tar.gz
+	sudo tar -C /usr/local -xzf go1.10.3.linux-amd64.tar.gz
+	popd
+    fi
+
+    export GOROOT=/usr/local/go
+    export GOPATH=~/go
+
+    go version
+}
 
 sudo apt-get install -y mtools dosfstools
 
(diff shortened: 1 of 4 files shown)


tools/setup-build-env.sh
```diff
@@ -2,11 +2,15 @@
 set -e
 set -x
 
-sudo sh -c 'echo "deb [arch=amd64] http://archive.ubuntu.com/ubuntu bionic main universe multiverse" > /etc/apt/sources.list'
-sudo sh -c 'echo "deb [arch=amd64] http://archive.ubuntu.com/ubuntu bionic-updates main universe multiverse" >> /etc/apt/sources.list'
-sudo sh -c 'echo "deb [arch=amd64] http://security.ubuntu.com/ubuntu bionic-security main universe multiverse" >> /etc/apt/sources.list'
+if [ -z "$OS_VERSION" ]; then
```
rbradford: Rather than add a new env variable - please use  `lsb_release -c  -s`

tools/CI/run_nats.sh
```diff
@@ -1,38 +1,50 @@
...
-
-pushd /tmp
-wget https://dl.google.com/go/go1.10.3.linux-amd64.tar.gz
-sudo tar -C /usr/local -xzf go1.10.3.linux-amd64.tar.gz
-popd
-
-export PATH=/usr/local/go/bin:$PATH
-export GOROOT=/usr/local/go
+GO_VERSION="1.10.3"
+CLEAR_VERSION=24740
```
rbradford, line 5: FYI, long term I think we need to find a better solution. Last week all old clear images (> 10 days) were purged and so everything broke in our CI system. I have some horrible go code (from Ciao) that parses the html to extract the latest cloud image. But am reluctant to use it...

tools/CI/run_nats.sh
```diff
@@ -1,38 +1,50 @@
...
+    fi
+
+    export GOROOT=/usr/local/go
+    export GOPATH=~/go
+
+    go version
+}
 
-go version
+sudo apt-get install -y mtools dosfstools
```
rbradford, line 26: I don't think your commit message is right for this commit. Nothing has changed with package downloading only with the image (and Go).

rbradford (approved):
lgtm with some small nits to fix first.

sameo:
@rbradford Updated, thanks.

PR #81, opened 2018-09-07
nats: Enable some basic tests with Ubuntu

rbradford:
Enable TestShutdown/TestReboot with Ubuntu as well as Clear.


rbradford:
Not quite ready to merge. Seems like the image fails to boot correctly (and so doesn't respond to SSH) unless you hack it to add "console=hvc0" to the grub command line.

This is something I did locally for testing and so when I ran this, everything passed, with a cleanly downloaded image SSH fails. I will review this next week...we might need to modify the image's grub config in the run_nats.sh script.

rbradford:
Xenial now working on some basic tests.

tools/CI/nats/nemu_test.go
```diff
@@ -317,51 +362,78 @@ func (q *qemuTest) startQemu(ctx context.Context, t *testing.T) error {
...
+				diskImage: d.image,
+				cloudInit: d.cloudInit,
+			}
+			ctx, cancelFunc := context.WithTimeout(context.Background(), cancelTimeout)
+			err := q.startQemu(ctx, t)
+			if err != nil {
+				cancelFunc()
+				<-q.doneCh
+				t.Fatalf("Error starting qemu: %v", err)
+			}
```
sameo, line 399: I suggest we put this block into its own routine as it's going to be repeated basically for each test that we want to run for each distro. But that's just a nit, this can be done as part of a follow up PR.

sameo (approved):

PR #83, opened 2018-09-10
hw: acpi: Don't put CPU and memory under PCI1 namespace

sboeuf:
There is no reason to put both CPU and memory hotplug under PCI1
namespace, but more than that, it's causing some trouble with PCI
experimentations when we're trying to allocate more than one PCI
bus.


sameo (approved):
LGTM

yangzhon:
LGTM

liujing2:
LGTM.
Tested cpu and memory hotplug and hot-unplug. Hot-unplug of memory still keeps the result and this patch doesn't break things.

PR #84, opened 2018-09-11
nats: Bump Clear version to 24950

rbradford:
Update to the latest Clear Linux release.


rbradford:
@sameo As expected hotplug is broken with latest Clear image

But there is also a kernel panic on shutdown - present on Q35/pc but not virt.

PR #89, opened 2018-09-13
hw: acpi: Fix the GED device definition

sboeuf:
Because the ACPI specification never mentions that a device such
as GED should be hosted under two _SB namespaces, this patch
simplifies its definition.


sameo:
LGTM

sameo (approved):

PR #90, opened 2018-09-14
build: Add Jenkinsfile entries to compile on ARM

rbradford:
Start adding support for CI building for ARM by using our newly configured
node.


rbradford:
I decided to go with the version using scripted syntax.

Issue #94, opened 2018-09-20
NEMU run some error


adofirst2018:
I try start NEMU follow this command， but  faced some error.

[root@duhy nemu]# qemu-system-x86_64 -m 2048M -smp 4 -drive file=qemu-3.0.qcow2,cache=none,if=virtio -machine virt,accel=kvm,kernel_irqchip -net nic -net tap,ifname=tap0,script=no,downscript=no
**
ERROR:qom/object.c:542:object_new_with_type: assertion failed: (type != NULL)
Aborted (core dumped)

[root@duhy nemu]# qemu-system-x86_64 -machine help
Supported machines are:
....
virt                 QEMU 3.0 i386 Virtual Machine (alias of virt-3.0)
virt-3.0             QEMU 3.0 i386 Virtual Machine
virt-2.12            QEMU 2.12 i386 Virtual Machine
isapc                ISA-only PC
none                 empty machine

adofirst2018:
(gdb) bt
#0  0x00007fffed82a277 in raise () at /lib64/libc.so.6
#1  0x00007fffed82b968 in abort () at /lib64/libc.so.6
#2  0x00007ffff74e00c5 in g_assertion_message () at /lib64/libglib-2.0.so.0
#3  0x00007ffff74e015a in g_assertion_message_expr () at /lib64/libglib-2.0.so.0
#4  0x0000555555cad188 in object_new_with_type (type=0x0) at qom/object.c:542
#5  0x0000555555cad200 in object_new (typename=0x0) at qom/object.c:556
#6  0x0000555555925a62 in cpu_new (typename=0x0, apic_id=0, errp=0x555556808cd8 <error_fatal>)
    at /home/duhy/workSpace/nemu/qemu/hw/i386/cpu.c:38
#7  0x00005555559261db in cpus_init (ms=0x55555691ba60, compat=false) at /home/duhy/workSpace/nemu/qemu/hw/i386/cpu.c:170
#8  0x00005555559527ac in virt_machine_state_init (machine=0x55555691ba60) at /home/duhy/workSpace/nemu/qemu/hw/i386/virt/virt.c:217
#9  0x0000555555aa316d in machine_run_board_init (machine=0x55555691ba60) at hw/core/machine.c:830
#10 0x00005555559e9593 in main (argc=13, argv=0x7fffffffe3a8, envp=0x7fffffffe418) at vl.c:4516

rbradford:
@duhongyu This won't happen if you add -cpu host

[...]

PR #96, opened 2018-09-20
i386/virt: Set default virt machine CPU type

rbradford:
The default CPU type for virt needs to be specified otherwise there is a
segfault if no CPU is specified on the command line.

### DIFF
--- a/hw/i386/virt/virt.c
+++ b/hw/i386/virt/virt.c
@@ -745,6 +745,7 @@ static void virt_machine_class_init(MachineClass *mc)
     mc->max_cpus = 288;
     mc->has_hotpluggable_cpus = true;
     mc->auto_enable_numa_with_memhp = true;
+    mc->default_cpu_type = X86_CPU_TYPE_NAME ("host");
 
     /* Machine class handlers */
     mc->cpu_index_to_instance_props = cpu_index_to_props;


hw/i386/virt/virt.c
```diff
@@ -745,6 +745,7 @@ static void virt_machine_class_init(MachineClass *mc)
     mc->max_cpus = 288;
     mc->has_hotpluggable_cpus = true;
     mc->auto_enable_numa_with_memhp = true;
+    mc->default_cpu_type = X86_CPU_TYPE_NAME ("host");
```
sboeuf, line 748: Aren't we hiding a potential bug by doing so?

rbradford: All the other machine types do this. What type of bug are you talking about? This was just missing in the virt machine.

We probably need to reject systems where the underlying cpu type is too old. But that's a separate problem to the failing to initialise the variable correctly.

sameo: I think @sboeuf means that we should not segfault on other cpu types, or if we do we ideally should have a way to specify we don't support it.
Do we know where the segfault was coming from ?

rbradford: @sameo @sboeuf 

If you look at the bt in #94 you'll see ms->cpu_type is passed unitialised into cpu_new(). This should never happen in QEMU as either a.) you've not specified a cpu type and the one from the mc->default_cpu_type will be used or b.) you've specified a cpu machine type.

This code has nothing to do with "other machine types" but instead "no machine type".

rbradford: Our problem was unlike the other machine types in QEMU, we forgot to set the default!

sameo: @rbradford I naively thought qemu was setting it to *some* default. It makes sense now, thanks.

sameo (approved):
LGTM

hw/i386/virt/virt.c
sboeuf: Yeah I get it now, thanks for the whole explanation :)

sboeuf (approved):

PR #103, opened 2018-09-27
default-configs: Simplify aarch64 build

rbradford:
Add minimial options needed to build for pure kvm ('virt') machine.


rbradford:
This is based off the configuration from experiment/code-reduction branch. I had to adapt it slightly as we now use some more features in NATS than we did then.

sameo (approved):
LGTM

Issue #102, opened 2018-09-27
The hw configurable impact other platform build 


yangzhon:
@sameo @rbradford 

I am disabling APIC and IOAPIC device model since the irq will offload to kernel.  when i did full build with below command
./configurable 

There are many build errors during all platform builds, some built-in files have been configurable, we need add this CONFIG_XXX to all platforms, this is the reason why i added one hw-common.mak before. If we do not use this file, we need to add each CONFIG_XXX into each XXX-softmmu.mak file because we are not sure if each platform will need this CONFIG_XXX. thanks!

PR #104, opened 2018-09-27
Make APIC and IOAPIC configurable in hw/intc

yangzhon:
Those two patches can make APIC and IOAPIC configurable, the virt platform will disable APIC and IOAPIC in qemu side because IRQ will offload to kernel side.

As for @sameo concern issue:  If we enable IRQ not in kernel mode, the virt platform will REPORT ASSERT error like below code
static void virt_ioapic_init(VirtMachineState *vms)
{
    qemu_irq *ioapic_irq;
    DeviceState *ioapic_dev;
    SysBusDevice *d;
    unsigned int i;

    /* KVM IRQ chip */
    assert(kvm_irqchip_in_kernel());
    ioapic_irq = g_new0(qemu_irq, IOAPIC_NUM_PINS);
    kvm_pc_setup_irq_routing(true);

With those two patches, the device model will down from 41 to 39, please @sameo @rbradford help review those patches, thanks!


rbradford:
Hi @yangzhon, there is a build error with your change:

```
/home/agentadmin/workspace/nemu_PR-104-I6ASOCKLYLC6PFFIKYBKLCG6WN4DPMKQ3WHCJGIWM5AMCVJJZKGQ/hw/intc/apic.c: In function ‘apic_sync_vapic’:
/home/agentadmin/workspace/nemu_PR-104-I6ASOCKLYLC6PFFIKYBKLCG6WN4DPMKQ3WHCJGIWM5AMCVJJZKGQ/hw/intc/apic.c:93:18: error: implicit declaration of function ‘get_highest_priority_int’ [-Werror=implicit-function-declaration]
         vector = get_highest_priority_int(s->isr);
                  ^
/home/agentadmin/workspace/nemu_PR-104-I6ASOCKLYLC6PFFIKYBKLCG6WN4DPMKQ3WHCJGIWM5AMCVJJZKGQ/hw/intc/apic.c:93:9: error: nested extern declaration of ‘get_highest_priority_int’ [-Werror=nested-externs]
         vector = get_highest_priority_int(s->isr);
         ^
```

yangzhon:
@rbradford 

thanks, i will check this after holiday.

By the way, please @rbradford @sameo help check issue  #102, find one way to resolve full build issue.
I felt there are some full build issue in this PR, so when i come back, i will firstly check #102  and then this PR. thanks!

yangzhon:
@rbradford @sameo 

Updated this PR, please help review, thanks.

sameo (approved):
LGTM

hw/intc/stub.c
```diff
@@ -23,3 +23,7 @@ int pic_read_irq(DeviceState *d)
 {
     return 0;
 }
+
+void ioapic_eoi_broadcast(int vector)
+{
+}
```
rbradford, line 30: I'm not sure always using a stub is a good idea. This function is only called in one place outside of the iopapic and apic files itself. So that could just be surrounded by #ifdef/#endif instead.

rbradford (approved):

hw/intc/stub.c
yangzhon: thanks, i have updated it.

yangzhon:
@rbradford @sameo 

Updated the code according to Rob's comments, please review again. thanks!

PR #106, opened 2018-09-27
tools: Reduce ARM64 build options

rbradford:
Reduce the configure flags for the ARM build to match those from the "pure
virt" build script in tools/build_x86_64_virt.sh


sameo (approved):
LGTM

sboeuf (approved):

rbradford:
@sameo looks like it's failing because of dependency on tcg header files (I think you fixed something like this recently?)

sameo:
@rbradford Let me look at that.

Issue #107, opened 2018-10-01
Ubuntu cloud images need console=hvc0


rbradford:
Currently in NATS and in our manual instructions we modify the the Ubuntu images to add console=hvc0 to the kernel command line.

This has the the benefit of giving us boot messages prior to getty.

_I think this is also required to workaround a bug where systemd doesn't boot correctly if ttyS0 cannot be correctly probed?_

Launchpad issue: https://bugs.launchpad.net/cloud-images/+bug/1795402

rbradford:
This project is no longer active and being archived. Cloud Hypervisor is the successor: https://github.com/cloud-hypervisor/cloud-hypervisor

PR #108, opened 2018-10-01
TESTING - See if image boots without guest image changes

rbradford:
Try and see if the image boots without modifying the guest image and without
trying to use the serial port or virtio console.


rbradford:
Seeing if my statement in #107 is still the case.

PR #111, opened 2018-10-02
i386: virt: Remove duplicate fields from VirtMachineState

rbradford:
Remove fields from VirtMachineState that are duplicates of fields that are
stored in AcpiConfiguration. This should prevent bugs that are the result of
field changes not be reflected between the different structs.

For brevity the member on VirtMachineState was also renamed and made an
embedded struct.

### DIFF
--- a/hw/i386/virt/memory.c
+++ b/hw/i386/virt/memory.c
@@ -42,7 +42,7 @@ MemoryRegion *virt_memory_init(VirtMachineState *vms)
     }
 
     vms->above_4g_mem_size = highmem_size;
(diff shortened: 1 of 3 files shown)


sboeuf (approved):

sboeuf:
Thanks @rbradford, this fixes the segfault issue I was having with nvdimm since the fields were not holding the right values.
@sameo the code is now preventing the duplication of the fields related to the `conf` structure, please take a look and merge if you're fine with this :)

sameo (approved):
LGTM

sameo:
Thanks @rbradford

PR #112, opened 2018-10-02
fw_cfg: Use fw_cfg to convey platform information

mcastelino:
Firmware (and OS) today uses the device id of the host bridge
as a means of determining the platform on which it is running.
Use fw_cfg instead to send this information. This will allow it
to be decoupled from PCI (as it should be) in the case of a
virtual platform.

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

The initialization is performed outside of `fw_cfg_init` as `fw_cfg_init` 
seems to be a little more generic. Or we can add this as a parameter
to fw_cfg_init.

### DIFF
--- a/include/hw/i386/pc.h
+++ b/include/hw/i386/pc.h
@@ -19,6 +19,16 @@
 #include "hw/mem/nvdimm.h"
 #include "hw/acpi/acpi_dev_interface.h"
 
+/* fw_cfg machine ids */
+enum {
+    X86_I440FX = 1,
+    X86_Q35,
+    X86_ISAPC,
+    X86_XENFV,
+    X86_XENPV,
+    X86_VIRT,
+};
+
 #define HPET_INTCAP "hpet-intcap"
 
 /**
--- a/hw/i386/virt/virt.c
+++ b/hw/i386/virt/virt.c
@@ -215,8 +215,8 @@ static void virt_machine_state_init(MachineState *machine)
                              "acpi-device", &error_abort);
             
     fw_cfg = fw_cfg_init(machine, smp_cpus, mc->possible_cpu_arch_ids(machine), vms->apic_id_limit);
+    fw_cfg_add_i16(fw_cfg, FW_CFG_MACHINE_ID, X86_VIRT);
     rom_set_fw(fw_cfg);
-
 
     if (machine->device_memory->base) {
         uint64_t *val = g_malloc(sizeof(*val));


mcastelino:
Related https://github.com/rbradford/edk2/pull/3

include/hw/i386/virt.h
```diff
@@ -53,6 +53,7 @@ typedef struct {
 } VirtMachineState;
 
 #define VIRT_MACHINE_NVDIMM "nvdimm"
+#define VIRT_QEMU_GPEX_DEVICE_ID 0x8
```
sameo: I guess `0x8` is coming from the PCI DID that the GPEX host currently uses?

sboeuf: Yeah I think so since this code only changes the way to pass this value, not the value itself.

sboeuf (approved):

include/hw/i386/virt.h
rbradford: As we're going to be moving away from using the GPEX DID for the PCI lite host bridge we should use this as an opportunity to use a better value for the machine type. It's important that we get this right as once we start upstreaming the OVMF changes we're not going to be able to modify it.

rbradford (changes requested):
I'm blocking this merge only because of needing to iterate on / discuss the value. The change itself is great and I think it is so much better than relying on PCI probing.

mcastelino:
> As we're going to be moving away from using the GPEX DID for the PCI lite host bridge we should use this as an opportunity to use a better value for the machine type. It's important that we get this right as once we start upstreaming the OVMF changes we're not going to be able to modify it.

@rbradford we should completely decouple the machine id from PCI DID. So we can use a machine id here and then OVMF can choose to map it to something else like the PCI DID. However machine id is not used consistently in QEMU.

The few examples I found are
```
enum cxmachines {
    CALXEDA_HIGHBANK,
    CALXEDA_MIDWAY,
};
```
```
enum {
    ss5_id = 32,
    vger_id,
    lx_id,
    ss4_id,
    scls_id,
    sbook_id,
    ss10_id = 64,
    ss20_id,
    ss600mp_id,
};
```

```
enum {
    ARCH_PREP = 0,
    ARCH_MAC99,
    ARCH_HEATHROW,
    ARCH_MAC99_U3,
};
```

Maybe we define an enum for x86 machine types?

rbradford:
@mcastelino Decoupling and using an enum gets a big :+1: from me

mcastelino:
@rbradford addressed your comments. Also https://github.com/rbradford/edk2/pull/3 has been updated to incorporate this change.

include/hw/i386/pc.h
```diff
@@ -19,6 +19,16 @@
...
+/* fw_cfg machine ids */
+enum {
+    X86_I440FX = 1,
+    X86_Q35,
+    X86_ISAPC,
+    X86_XENFV,
+    X86_XENPV,
+    X86_VIRT,
+};
+
```
sameo, line 31: That looks good to me, but could we move that enum either under a new `include/hw/i386/machine.h` file or under the already existing `include/hw/i386/fw.h` one?

PR #113, opened 2018-10-02
acpi: Do not hardcode the segment number and PCI Hotplug IO Port Address

mcastelino:
Do not hardcode the namespace object associated with the host bridge
and _UID. Explicitly set the segment number in the ACPI table.

Specify ACPI PCI Hotplug IO Port address as a parameter. This will
allow each platform to choose its own memory layout.


hw/acpi/aml-build.c
```diff
@@ -2189,7 +2189,7 @@ Aml *build_prt(bool is_pci0_prt)
     return method;
 }
 
-Aml *build_pci_host_bridge(Aml *table, AcpiPciBus *pci_host)
+Aml *build_pci_host_bridge(Aml *table, AcpiPciBus *pci_host, uint16_t pcie_segment)
```
sboeuf: The commit looks good to me, but just a random thought here. Why don't we modify the `AcpiPciBus` structure by adding a new field `index`, instead of passing the PCI segment number?

mcastelino: That is indeed a better way to do it. That way we can pass more information in the future.

mcastelino: @sboeuf fixed..

include/hw/acpi/aml-build.h
```diff
@@ -235,6 +235,7 @@ typedef struct AcpiPciBus {
     PCIBus *pci_bus;
     Range *pci_hole;
     Range *pci_hole64;
+    uint16_t pci_segment;
```
sboeuf: Maybe it'd be worth putting a comment here explaining that `uint16` was chosen on purpose as we can support 65536 buses, is that it?

hw/acpi/aml-build.c
```diff
@@ -2317,26 +2317,26 @@ Aml *build_pci_host_bridge(Aml *table, AcpiPciBus *pci_host)
...
-void build_acpi_pci_hotplug(Aml *scope)
+void build_acpi_pci_hotplug(Aml *scope, uint16_t acpi_iobase_addr)
 {
     Aml *field;
     Aml *method;
 
-    aml_append(scope,
-        aml_operation_region("PCST", AML_SYSTEM_IO, aml_int(0xae00), 0x08));
+    aml_append(scope, aml_operation_region("PCST", AML_SYSTEM_IO, 
+                              aml_int(acpi_iobase_addr), 0x08));
```
sboeuf: Just a side note, but I think the size of `PCST`, `SEJ`, and `BNMR` could be held by pre-defined constants such as:
```c
#define ACPI_PCST_LEN 0x08
...
```

include/hw/acpi/pcihp.h
```diff
@@ -43,6 +43,13 @@ typedef struct AcpiPciHpPciStatus {
 #define ACPI_PCIHP_MAX_HOTPLUG_BUS 256
 #define ACPI_PCIHP_BSEL_DEFAULT 0x0
 
+#define ACPI_PCIHP_SIZE 0x0014
+#define ACPI_PCI_UP_BASE 0x0000
+#define ACPI_PCI_DOWN_BASE 0x0004
+#define ACPI_PCI_EJ_BASE 0x0008
```
sboeuf, line 52: I appreciate the addition of `ACPI_` prefix for those constants, it makes things clearer.

include/hw/acpi/pc-hotplug.h
```diff
@@ -32,4 +32,7 @@
 
 #define ACPI_MEMORY_HOTPLUG_BASE 0x0a00
 
+#define PIIX_ACPI_PCI_HOTPLUG_IO_BASE 0xae00
+#define VIRT_ACPI_PCI_HOTPLUG_IO_BASE 0xae00
```
sboeuf, line 36: I'm fine with using the same base address as `piix` as the base address, but what's the rationale behind it if we get asked the question: _Why `piix` based address and not `ich9` one?_

sboeuf (approved):
Globally looks good :)

include/hw/acpi/aml-build.h
mcastelino: This is actually the `PCI Segment Group Number`. I will add that detail.

PCI Conventional was limited to a maximum of 256 PCI bus segments. PCI Express extends this by introducing "PCI Segment Groups", where a system could (in theory) have up to 65536 PCI Segment Groups with 256 PCI bus segments per group, thereby allowing a single computer to have up to a maximum of 16777216 PCI bus segments.

From the PCI Firmware Spec.



| Offset  | Length | Description|
|---------| ------- |--------------|
|0 | 4 | Table Signature ("MCFG")|
|4 | 4 | Length of table (in bytes)|
|8 | 1 | Revision (1)|
|9 | 1 | Checksum (sum of all bytes in table & 0xFF = 0)|
|10 | 6 | OEM ID (same meaning as other ACPI tables)|
|16 | 8 | OEM table ID (manufacturer model ID)|
|24 | 4 | OEM Revision (same meaning as other ACPI tables)|
|28 | 4 | Creator ID (same meaning as other ACPI tables)|
|32 | 4 | Creator Revision (same meaning as other ACPI tables)|
|36 | 8 | Reserved|
|44 | 16 * n | Configuration space base address allocation structures. Each structure uses the following format:|
|0 | 8 | Base address of enhanced configuration mechanism|
|8 | 2 | PCI Segment Group Number|
|10 | 1 | Start PCI bus number decoded by this host bridge|
|11 | 1 | End PCI bus number decoded by this host bridge
|12 | 4 | Reserved

hw/acpi/aml-build.c
mcastelino: Makes sense. Fewer manifest constants the better.

include/hw/acpi/pc-hotplug.h
mcastelino: @sboeuf in a future commit I plan to completely define the IO and Memory maps for the virt platforms in its own file along the lines of https://github.com/intel/nemu/blob/topic/virt-x86/hw/arm/virt.c#L126 but not in exactly the same manner. Also it may allow us to use the IO space more efficiently.

That ensures that at least for the virt platform we know exactly all the IRQs, IO Ports and Memory that we are using. OVMF adds a few things on top of this. But this map will be definitive at least from the QEMU point of view.

For now we chose the piix as Q35 does not have this defined AFAIK as it does not support ACPI based PCI hotplug.

sboeuf: Ok that's a very clear answer, I just wanted to make sure we had some justification here.

I think the code you pointed out is a perfect example as we can understand the whole layout of virt, just looking at a single file. This will help tidying things up :)

mcastelino:
@sboeuf comments addressed.

include/hw/acpi/aml-build.h
sboeuf: I cannot find where you added this detail?

sboeuf (approved):

PR #115, opened 2018-10-03
config: Add PCI Bridge and PCIe bridge support as Kata uses them

mcastelino:
config: Add PCI Bridge and PCIe bridge support as Kata uses them

Add PCIe bridge support to virt as Kata uses them. Kata does not
need PCIe Root Ports or SHPC with virt as it uses ACPI hotplug.
This brings in SHPC, as SHPC can only be disabled at runtime today.

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>


sboeuf (approved):

sameo (approved):
LGTM

mcastelino:
Note: Kata uses a pcie-pci-bridge for now to avoid having to change the Kata codebase across Q35 and virt. Longer term once we change the Kata code we can directly hotplug devices on bus 0 using APCI which was not possible in Q35.

PR #119, opened 2018-10-04
ACPI and PCI Changes

mcastelino:
- Use fw_cfg to send in machine_id
- Do not hardcode PCI segment number or ACPI Port Addresses


sboeuf (approved):

sameo (approved):
LGTM

mcastelino:
Fails with updated OVMF https://github.com/intel/nemu/issues/120
This is a latent bug that surfaced when virt was exposed as a PCIe machine type.

PR #128, opened 2018-10-09
DMI: set default values for SMBIOS (If we need DMI)

liujing2:
@mcastelino @yangzhon 

If we want to enable DMI, we need firstly set some default values for
SMBIOS, otherwise, smbios_get_tables() will return NULL directly for
the two tables and guest will show as follows.

root@clr-19bebbac82a145bb8c96500c89dbfd8a ~ # dmesg | grep DMI
[    0.000000] DMI not present or invalid.

This makes things as follows, on virt platform. BTW, simiply remove
the CONFIG_SMBIOS check in fw_build_smbios() for now because it does not
give an expective value of CONFIG_SMBIOS.

root@clr-19bebbac82a145bb8c96500c89dbfd8a ~ # dmesg|grep DMI
[    0.000000] DMI: QEMU VIRTUAL MACHINE(virt), BIOS 0.0.0 02/06/2015

### DIFF
--- a/hw/i386/fw.c
+++ b/hw/i386/fw.c
@@ -34,7 +34,11 @@ struct hpet_fw_config hpet_cfg = {.count = UINT8_MAX};
 
 static void fw_build_smbios(MachineState *ms, FWCfgState *fw_cfg)
 {
-#if defined(CONFIG_SMBIOS)
+/* TODO: This does not work as expect because CONFIG_SMBIOS=y is
+ * configured inside build/hw/smbios? While fw.o is under x86_64-softmmu/.
+ * To temporarily enable SMBIOS, remove the check for now
+ */
+//#if defined(CONFIG_SMBIOS)
     uint8_t *smbios_tables, *smbios_anchor;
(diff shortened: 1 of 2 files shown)


yangzhon:
If we mask #if defined(CONFIG_SMBIOS), the pure virt build will fail.

I will look for another solution to compile SMBIOS out.

liujing2:
@sboeuf What information is in the minimal DMI?

sboeuf:
@liujing2 as @mcastelino explained it to me, the DMI can be used to pass a lot of informations, but in our case, those informations are already passed through other ways, that's why we thought we didn't need to pass a DMI table at all.

But the guest Linux kernel will do one strong assumption if no DMI table is passed, which is that the BIOS must be very old. When this assumption is made, the kernel will not accept MCFG table to be written above 4G.

The simple way to let the kernel think that the platform is recent enough is to provide a date more recent than `January 2010` (@mcastelino correct me if I'm wrong here), which means the DMI table only needs to provide this information.

PR #130, opened 2018-10-09
i386: virt: Remove emulated IOMMUs from allowed devices

rbradford:
Remove the emulated IOMMUs from the list of devices that the user is
allowed to create when using the virt platform.


sameo:
I am fine with that one as I'm not sure why we'd want to emulate an IOMMU from virt. But I may be missing some use cases and I'd like to get @mcastelino input.

rbradford:
And further, they don't work as they assume PC machines.

sameo:
> And further, they don't work as they assume PC machines.

Correct. From `hw/i386/x86-iommu.c` for example:

```C
static void x86_iommu_realize(DeviceState *dev, Error **errp)
{
    X86IOMMUState *x86_iommu = X86_IOMMU_DEVICE(dev);
    X86IOMMUClass *x86_class = X86_IOMMU_GET_CLASS(dev);
    MachineState *ms = MACHINE(qdev_get_machine());
    MachineClass *mc = MACHINE_GET_CLASS(ms);
    PCMachineState *pcms =
        PC_MACHINE(object_dynamic_cast(OBJECT(ms), TYPE_PC_MACHINE));
    QLIST_INIT(&x86_iommu->iec_notifiers);

    if (!pcms || !pcms->bus) {
        error_setg(errp, "Machine-type '%s' not supported by IOMMU",
                   mc->name);
        return;
    }
[...]
}
```

That being said, I think the x86 IOMMU drivers for QEMU are wrongly designed. It currently depends on PC machines because it needs a PCI bus, there are no PC machine type specific dependency per se. 
However, the `x86_iommu` type's parent is the system bus whereas having it being a PCI bus child would make it completely PC machine independent for it to fetch a PCI bus.

So, if we decide that we still want to support IOMMU emulation, we need to fix that. Otherwise, this patch LGTM.

sameo:
@rbradford Also, could we disable `CONFIG_VTD` and `CONFIG_AMD_IOMMU` from the virt config?

rbradford:
> @rbradford Also, could we disable `CONFIG_VTD` and `CONFIG_AMD_IOMMU` from the virt config?

Great idea. Done!

sboeuf (approved):

mcastelino (approved):
LGTM.

If we do want to support it we will need to support at the very minimum

```
kernel_irqchip=split
```

mcastelino:
> I am fine with that one as I'm not sure why we'd want to emulate an IOMMU from virt. But I may be missing some use cases and I'd like to get @mcastelino input.

The place where we may need it is the case of vfio. When we directly assign a device to the VM, and we want to offer VT-d protections to the guest memory.

sameo (approved):
LGTM

Issue #127, opened 2018-10-09
Add nemu-specific tag to "--version" output


jodh-intel:
To allow users to determine if their qemu binary is in fact NEMU, please add a unique tag (`nemu`? :-) to the output of `--version`.

This will be particularly useful when [Kata Containers are configured to use NEMU](https://github.com/kata-containers/documentation/pull/266) since running `kata-runtime kata-env` will then capture this important detail to determine which version of QEMU is being used regardless of the binary name.

/cc @sboeuf.

sboeuf:
Thanks @jodh-intel
@sameo @rbradford what do you think about this? This would make nemu identification clear.

sameo:
PR #131 fixed that one.

PR #131, opened 2018-10-09
vl: Adjust QEMU system binary to indicate this is NEMU

rbradford:
For debug reports indicate that this is a NEMU binary not a QEMU one.


jodh-intel:
lgtm - this will fix #127.

sameo (approved):

PR #139, opened 2018-10-15
nats: Update for govmm changes

rbradford:
The type signature for qmp.ExecuteCPUDeviceAdd() has changed in order to
support supplying a path to a ROM file to use. As we are using OVMF we
can supply an empty romfile.


rbradford:
@sameo long term we can look at some vendoring but i'm not current on what the best technique is now.

sboeuf (approved):

PR #140, opened 2018-10-16
hw: acpi: Multiplex GED events on a single interrupt

sboeuf:
This patch allows for the definition of a single interrupt for the
GED device, relying on IO memory region to communicate the type of
device affected by the interrupt.

This way, we can support up to 2^32 events with a unique interrupt.

Here is the example of what we should expect from the DSDT table with
those new changes:
```c
    Device (\_SB.GED)
    {
        Name (_HID, "ACPI0013" /* Generic Event Device */)  // _HID: Hardware ID
        Name (_UID, Zero)  // _UID: Unique ID
        Name (_CRS, ResourceTemplate ()  // _CRS: Current Resource Settings
        {
            Interrupt (ResourceConsumer, Level, ActiveHigh, Exclusive, ,, )
            {
                0x00000010,
            }
        })
        OperationRegion (IREG, SystemIO, 0xB000, 0x04)
        Field (IREG, DWordAcc, NoLock, WriteAsZeros)
        {
            ISEL,   32
        }

        Method (_EVT, 1, Serialized)  // _EVT: Event
        {
            Local0 = ISEL /* \_SB_.GED_.ISEL */
            If (((Local0 & One) == One))
            {
                \_SB.CPUS.CSCN ()
            }

            If (((Local0 & 0x02) == 0x02))
            {
                \_SB.MHPC.MSCN ()
            }

            If (((Local0 & 0x04) == 0x04))
            {
                Notify (\_SB.NVDR, 0x80) // Status Change
            }

            If (((Local0 & 0x08) == 0x08))
            {
                Acquire (\_SB.PCI0.BLCK, 0xFFFF)
                \_SB.PCI0.PCNT ()
                Release (\_SB.PCI0.BLCK)
            }
        }
    }
```


sboeuf:
FYI, I have tested with PCI and CPU hotplug, and it works!

rbradford:
Please add a PCI hotplug to NATS to test this.

include/hw/acpi/pc-hotplug.h
```diff
@@ -35,4 +35,6 @@
 #define PIIX_ACPI_PCI_HOTPLUG_IO_BASE 0xae00
 #define VIRT_ACPI_PCI_HOTPLUG_IO_BASE 0xae00
 
+#define ACPI_GED_EVENT_IO_BASE 0xb000
```
mcastelino: @sboeuf can this instead be something that the virt platform definition controls and setups at 

```
    acpi_ged_init(get_system_io(), OBJECT(dev),
                  &s->ged_state, VIRT_ACPI_GED_EVENT_IO_BASE);
```
vs being a constant. Even if it is a constant can it come from the virt platform .h file ?

I plan to do the same for the other PCI and other hotplug defines too. This ensures that a platform is self contained.

hw/i386/virt/virt.c
```diff
@@ -96,19 +97,19 @@ static void acpi_conf_virt_init(MachineState *machine)
     /* GED events */
     GedEvent events[] = {
         {
-            .irq   = VIRT_GED_CPU_HOTPLUG_IRQ,
+            .irq   = ACPI_GED_IRQ_SEL_CPU,
```
mcastelino: Why are we defining multiple IRQs here?

include/hw/acpi/ged.h
```diff
@@ -0,0 +1,46 @@
...
+#ifndef HW_ACPI_GED_H
+#define HW_ACPI_GED_H
+
+#include "qemu/osdep.h"
+#include "exec/memory.h"
+
+#define ACPI_GED_IRQ_SEL_OFFSET 0x0
+#define ACPI_GED_IRQ_SEL_LEN    0x4
+#define ACPI_GED_IRQ            0x10
+#define ACPI_GED_IRQ_SEL_CPU    0x0
```
mcastelino: Why are we defining multiple IRQs here?

include/hw/acpi/ged.h
```diff
@@ -0,0 +1,46 @@
...
+
+#ifndef HW_ACPI_GED_H
+#define HW_ACPI_GED_H
+
+#include "qemu/osdep.h"
+#include "exec/memory.h"
+
+#define ACPI_GED_IRQ_SEL_OFFSET 0x0
+#define ACPI_GED_IRQ_SEL_LEN    0x4
+#define ACPI_GED_IRQ            0x10
```
mcastelino: `ACPI_GED_IRQ` should ideally be chosen by the platform and not implicit in the GED definition itself. That will make this code generic.

Also in the future if we need to support multiple IRQs for the GED device we should probably adopt a notion of IRQ base and offset like how we do for the op regions. This will make the placements more generic.

hw/acpi/ged.c
```diff
@@ -0,0 +1,70 @@
...
+
+void acpi_ged_init(MemoryRegion *as, Object *owner,
+                   GEDState *state, hwaddr base_addr)
+{
+    memory_region_init_io(&state->io, owner, &ged_ops, state,
+                          "acpi-ged-event", ACPI_GED_REG_LEN);
+    memory_region_add_subregion(as, base_addr, &state->io);
+}
+
+void acpi_ged_event(GEDState *ged_st, qemu_irq *irq, uint32_t ged_irq_sel)
```
mcastelino, line 66: Does it make sense for ACPI_GED_IRQ to be passed in as a parameter. That way it is driven from the platform hotplug routine.

mcastelino (commented):
@sboeuf overall it looks good and simplifies the hotplug logic and may even allow us to leverage this for sending in a unique callback PCI domain.

Left feedback primarily around making the device more generic and allowing the platform to control the memory and IRQ maps. This will make this code reusable on other platforms.

include/hw/acpi/pc-hotplug.h
sboeuf: I like that, I can definitely introduce this change!

hw/i386/virt/virt.c
sboeuf: Well we're not. I kept the name `irq`, but it's more about the selector value related to each use case. Maybe I should change the `irq` into `selector` or `offset`.
@sameo @mcastelino wdyt?

include/hw/acpi/ged.h
sboeuf: We're not, they are the value of the selector for each type of interrupt.
But there's only one IRQ, which is:
```c
#define ACPI_GED_IRQ 0x10
```

include/hw/acpi/ged.h
sboeuf: >`ACPI_GED_IRQ` should ideally be chosen by the platform and not implicit in the GED definition itself. That will make this code generic.

Yes I agree on this one. Maybe it could be passed as part of the init `acpi_ged_init()`?

>Also in the future if we need to support multiple IRQs for the GED device we should probably adopt a notion of IRQ base and offset like how we do for the op regions. This will make the placements more generic.

I need your input on this before to proceed further because I simplified the GED model by limiting it to a single and unique interrupt, since the 32bits selector is large enough to handle 2^32 use cases. /cc @sameo

hw/acpi/ged.c
sboeuf: Yes exactly ;)

sboeuf:
@mcastelino 
>@sboeuf overall it looks good and simplifies the hotplug logic and may even allow us to leverage this for sending in a unique callback PCI domain.

Do you mean MSI? Not sure I understand the PCI usage.
It could be used for MSI domain if we can specify a domain of one single interrupt :)

include/hw/acpi/ged.h
mcastelino: @sboeuf yes for our platform we want a single interrupt. For other platforms they might want unique interrupts. All I was saying is allow the flexibility for the platform to choose one or more interrupts. The ACPI table generation is controlled by the platform. Also the hotplug logic is again platform controlled. Hence allow the platform to choose the IRQ (or IRQs) and the Port address.

hw/acpi/ged.c
mcastelino: @sboeuf so you will pass both the selector and the IRQ as parameters?

hw/i386/virt/virt.c
mcastelino: Where is .irq used now?

include/hw/acpi/ged.h
mcastelino: You are specifying it in a ged specific file and platform specific. I was suggesting that you not define it here and allow the platform to define it.

hw/i386/virt/virt.c
sboeuf: It's used from `hw/acpi/reduced.c` by the function `build_ged_aml()` building the GED device definition for the DSDT table.

include/hw/acpi/ged.h
sboeuf: Yes I agree we can let the platform define `ACPI_GED_IRQ`, but there's no point in letting the platform defining `ACPI_GED_IRQ_SEL_CPU`, `ACPI_GED_IRQ_SEL_MEM` and so on, right?

include/hw/acpi/ged.h
sboeuf: If we introduce the support for multiple IRQs for GED, I'll need to change the code deeply (which is fine if that's what we want). I'm just wondering why another platform would want to support multiple IRQs since the multiplexing covers a huge amount of cases already?

What is the `Port address` that you mentioned?

hw/acpi/ged.c
sboeuf: Sorry, I meant I was planning to pass `ACPI_GED_IRQ` as a parameter through `acpi_ged_init()`

hw/i386/virt/virt.c
sameo: > Well we're not. I kept the name `irq`, but it's more about the selector value related to each use case. Maybe I should change the `irq` into `selector` or `offset`.
> @sameo @mcastelino wdyt?

selector makes more sense, yes,

hw/acpi/Makefile.objs
```diff
@@ -7,6 +7,7 @@ common-obj-$(CONFIG_ACPI_CPU_HOTPLUG) += cpu_hotplug.o
 common-obj-$(CONFIG_ACPI_MEMORY_HOTPLUG) += memory_hotplug.o
 common-obj-$(CONFIG_ACPI_CPU_HOTPLUG) += cpu.o
 common-obj-$(CONFIG_ACPI_NVDIMM) += nvdimm.o
+common-obj-$(CONFIG_ACPI_GED) += ged.o
```
sameo: Is there any reason why we're defining a new config for that one? Can't we simplify by adding ged.o to `ACPI_HW_REDUCED` ?

include/hw/acpi/pc-hotplug.h
sameo: It would be good for `acpi_ged_init()` to take the GED IRQ and the IO base as platform defined arguments.

include/hw/acpi/ged.h
sameo: @mcastelino I think providing support for both unique and multiple interrupts would make both the implementation and the API more complex. AFAICT, the only pro for having multiple interrupts would be latency but I don't think it matters that much for GED use cases.
I prefer for the GED implementation to handle the interrupt multiplexing internally and provide an event based API, where platforms only define the kind of events they want to support through GED. This PR goes into that direction.

include/hw/acpi/ged.h
sameo: > Yes I agree we can let the platform define `ACPI_GED_IRQ`, but there's no point in letting the platform defining `ACPI_GED_IRQ_SEL_CPU`, `ACPI_GED_IRQ_SEL_MEM` and so on, right?

I think you only want the platform to specify the IO base for the memory region.

sameo (changes requested):

sboeuf:
@rbradford https://github.com/intel/nemu/pull/141 adds a PCI HP test.

hw/acpi/Makefile.objs
mcastelino: @sameo yes just the base

sboeuf: Oh yes that's better, I'll do it.

mcastelino: But does that not imply that we will always need to use/define the GED device. If we do not need to support runtime hotplug, then we do not need to support the GED device.

sboeuf: I didn't think about that, but yes, it would allow for a finer granularity.

sboeuf:
@mcastelino @sameo comments addressed. We need to decide what we want to do about `CONFIG_ACPI_GED` vs `CONFIG_ACPI_HW_REDUCED` ?

mcastelino:
> @mcastelino @sameo comments addressed. We need to decide what we want to do about `CONFIG_ACPI_GED` vs `CONFIG_ACPI_HW_REDUCED` ?

@sboeuf I am ok with CONFIG_ACPI_HW_REDUCED for now. Once we are able to cleanly support a model where we do not need hotplug (or at least runtime hotplug) we can add this.

hw/acpi/ged.c
```diff
@@ -0,0 +1,223 @@
...
+    switch (event->event) {
+    case GED_CPU_HOTPLUG:
+        /* We run a complete CPU SCAN when getting a CPU hotplug event */
+        return aml_call0("\\_SB.CPUS." CPU_SCAN_METHOD);
+    case GED_MEMORY_HOTPLUG:
+        /* We run a complete memory SCAN when getting a memory hotplug event */
+        return aml_call0("\\_SB.MHPC." MEMORY_SLOT_SCAN_METHOD);
+    case GED_PCI_HOTPLUG:
+	/* Take the PCI lock and trigger a PCI rescan */
+        method = aml_acquire(aml_name("\\_SB.PCI0.BLCK"), 0xFFFF);
```
mcastelino, line 97: @sboeuf not directly related to this, but as the method itself is serialized, do we need the lock?

hw/acpi/cpu.c
```diff
@@ -293,23 +293,6 @@ const VMStateDescription vmstate_cpu_hotplug = {
     }
 };
 
-#define CPU_NAME_FMT      "C%.03X"
```
mcastelino, line 296: @sboeuf minor nit. We could have made these moves in a different commit unless I am missing something.

hw/acpi/cpu_hotplug.c
```diff
@@ -169,11 +162,11 @@ void build_legacy_cpu_hotplug_aml(Aml *ctx, MachineState *machine,
     aml_append(method, else_ctx);
     aml_append(sb_scope, method);
 
-    method = aml_method(CPU_EJECT_METHOD, 2, AML_NOTSERIALIZED);
+    method = aml_method(CPU_LEGACY_EJECT_METHOD, 2, AML_NOTSERIALIZED);
```
mcastelino, line 165: @sboeuf maybe I am missing something. Why this change. The commit message does not help here.

mcastelino (commented):
@sboeuf changes look good. However if it is not too much rework this should be split into two commits. The GED specific changes and the other code motion around ACPI code motion.

hw/acpi/ged.c
sboeuf: I would say no. There's no reason for locking IMO.

hw/acpi/cpu.c
sboeuf: definitely!

hw/acpi/cpu_hotplug.c
sboeuf: I will make this change part of the rework with the new commit. The idea being to avoid the problem of having the same constants `CPU_EJECT_METHOD` and `CPU_SCAN_METHOD` being defined both from `include/hw/acpi/cpu.h` and `include/hw/acpi/cpu_hotplug.h`

sboeuf:
@mcastelino 
>@sboeuf I am ok with CONFIG_ACPI_HW_REDUCED for now. Once we are able to cleanly support a model where we do not need hotplug (or at least runtime hotplug) we can add this.

This makes sense, let's go step by step here.

>@sboeuf changes look good. However if it is not too much rework this should be split into two commits. The GED specific changes and the other code motion around ACPI code motion.

Yes, I'll try to organize through different commits to help the understanding of what's being done here.

sboeuf:
@rbradford same CI issue:
```
[nemu_PR-141-63PF7XB72QX7OHRMVS6OUPQJZ7CZWGTST6OIDKN5TPTY763NOADA] Running shell script
+ sudo apt-get build-dep -y qemu
E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)
E: Unable to lock the administration directory (/var/lib/dpkg/), is another process using it?
```

hw/acpi/ged.c
```diff
@@ -0,0 +1,223 @@
...
+    memory_region_add_subregion(as, base_addr, &ged_st->io);
+}
+
+void acpi_ged_event(GEDState *ged_st, qemu_irq *irq, uint32_t ged_irq_sel)
+{
+    /* Set the GED IRQ selector to the expected device type value. This
+     * way, the ACPI method will be able to trigger the right code based
+     * on a unique IRQ.
+     */
+    ged_st->sel = ged_irq_sel;
```
sameo: So what happens if I trigger a CPU hotplug and then a memory hotplug before the guest has read the first hotplug event? In other words, to avoid this to be racy, shouldn't we use a bitmask here?

sameo (changes requested):

sboeuf:
@mcastelino @sameo I have addressed your comments, please let me know how it looks.

sboeuf:
@sameo about bitmask, I have chosen to write back to the memory IO region in order to reset only the bit affected by the interrupt. This way, each interrupt (even in case of two simultaneous hotplugs) will result in hotplugging the appropriate device type.

hw/acpi/ged.c
```diff
@@ -17,6 +17,69 @@
...
+        val = ged_st->sel;
+        break;
+    default:
+        break;
+    }
+
+    return val;
+}
+
+/* Nothing is expected to be written to the GED memory region */
```
sameo, line 40: Please remove that comment.

hw/acpi/ged.c
```diff
@@ -67,90 +133,82 @@ void build_ged_aml(Aml *table, const char *name,
...
-         *        MethodEvent1()
-         *        Local0 = Zero
-         *    }
-         *    ElseIf (Arg0 == irq2)
-         *    {
-         *        MethodEvent2()
-         *        Local0 = Zero
-         *    }
+         *     MethodEvent0()
+         *     ISEL = irq0
```
sameo: I think something like that would be clearer:

```
ISEL = Local0 & ~irq0
```

Otherwise, reading the AML code without looking at the corresponding QEMU code is confusing.

sameo (changes requested):
Thanks for the changes, almost here... I have one nit left, and one AML generation code change request.

sameo (approved):
LGTM

sameo:
if @mcastelino is fine with that one, we can merge.

include/hw/acpi/cpu.h
```diff
@@ -17,6 +17,23 @@
 #include "hw/acpi/aml-build.h"
 #include "hw/hotplug.h"
 
+#define CPU_NAME_FMT      "C%.03X"
```
mcastelino, line 20: @sboeuf is there any reason these cannot be in cpu_hotplug.h vs here? memory hotplug ACPI defines all sit nicely in a single file.

sboeuf: Oh yeah we could go one step further and move everything to a single file I think. Just didn't want to go that far, but your call, I can make the change if you think it's worth it.

mcastelino: No need for that. I was wondering if it had anything to do with 
`#define CPU_EJECT_METHOD  "CEJ0"` being here and `#define CPU_LEGACY_EJECT_METHOD "CPEJ"` being in the other file.

mcastelino (approved):
LGTM

rbradford:
This needs a rebase before it can be merged due to removed changes on the branch.

PR #142, opened 2018-10-17
acpi: reduced: Remove all virt machine type dependencies

sameo:
We are directly fetching the PCI bus from a virt machine type specific
macro, but there are not technical reasons why the ACPI hardware-reduced
implementation should be linked to the virt machine type.
We fetch the PCI bus through the ACPI instead, making this
implementation machine type agnostic.


yangzhon:
LGTM, thanks!

rbradford (approved):

PR #143, opened 2018-10-17
i386: virt: Make BIOS region RAM rather than ROM

rbradford:
On other machine types, such as Q35, the PAM registers on the PMC can be
used to make this region read/write. However in virt we do not have that
capability. My making the BIOS region read/write we can support simpler
firmware solutions such as qboot.

### DIFF
--- a/hw/i386/virt/memory.c
+++ b/hw/i386/virt/memory.c
@@ -99,7 +99,7 @@ MemoryRegion *virt_memory_init(VirtMachineState *vms)
                                     &machine->device_memory->mr);
     }
 
-    sysfw_firmware_init(system_memory, false);
+    sysfw_firmware_init(system_memory, true);
     
     return ram;
 }


hw/i386/virt/memory.c
```diff
@@ -99,7 +99,7 @@ MemoryRegion *virt_memory_init(VirtMachineState *vms)
                                     &machine->device_memory->mr);
     }
 
-    sysfw_firmware_init(system_memory, false);
+    sysfw_firmware_init(system_memory, true);
```
sboeuf, line 102: By doing so, `sysfw_firmware_init()` calls into `old_pc_system_rom_init()`. Do we actually want to execute the full function here, or should we factorize only the part that we need?

sameo (approved):
LGTM

hw/i386/virt/memory.c
rbradford: @sboeuf  We call into `old_pc_system_rom_init()` anyway as we don't have a pflash to boot from.

sboeuf: Nevermind, I don't know how to read the code....
Your change is only preventing the memory region to be set `readonly`.

sboeuf (approved):

PR #146, opened 2018-10-18
Add a debugexit device on the system bus

sameo:
For non legacy x86 machine types (e.g. virt) that do not support ISA,
    the isa-debug-exit device does not work.
    This is the equivalent of the ISA debugexit device, but running off the
    system bus.
    
With that device, adding e.g. `-device sysbus-debugexit,iobase=0xf4` to
    the QEMU command line will terminate the QEMU process, which is a very
    convenient way to measure specific latencies with basic tools like
    `time`.


rbradford (approved):

sboeuf (approved):
Nice!

PR #156, opened 2018-10-24
accel: use kvm as default

terenceli:
As we are for the cloud, it makes sense to use kvm as default accelerator.
Without this patch, the CLI "-machine virt" will complain it can't find the
TCG accelerator.


accel/accel.c
```diff
@@ -79,8 +79,8 @@ void configure_accelerator(MachineState *ms)
 
     accel = qemu_opt_get(qemu_get_machine_opts(), "accel");
     if (accel == NULL) {
-        /* Use the default "accelerator", tcg */
-        accel = "tcg";
+        /* Use the default "accelerator", kvm */
+        accel = "kvm";
```
sameo, line 83: @terenceli thanks for the patch.
This is going to impact all machine types, for all use cases, and is going to be a hard upstream sell. Could you please try to make it `i386/virt` specific by setting the accel property to `kvm` if it's empty?

terenceli: Right,
Until the discussion in issue [157](https://github.com/intel/nemu/issues/157), I realized it is not that easy to got involved as every change will need to be considered with qemu. I will see more nemu code and resent this patch.

Thanks,

rbradford:

PR #158, opened 2018-10-25
Remove isa-bios memory region in virt platform

yangzhon:
When i use "info qom-tree" command and find "/isa-bios[0] (qemu:memory-region)" in memory region list.

The isa-bios is not useful in virt platform and i added one bool parameter to check this.

By the way, this patch also fixed another pflash issue in x86.
Memory.c (hw\i386\virt):    sysfw_firmware_init(system_memory, true);

void sysfw_firmware_init(MemoryRegion *rom_memory, bool isapc_ram_fw)
{
    DriveInfo *pflash_drv;

    pflash_drv = drive_get(IF_PFLASH, 0, 0);

    if (isapc_ram_fw || pflash_drv == NULL) {
        /* When a pflash drive is not found, use rom-mode */
        old_pc_system_rom_init(rom_memory, isapc_ram_fw);
        return;
    }

    if (kvm_enabled() && !kvm_readonly_mem_enabled()) {
        /* Older KVM cannot execute from device memory. So, flash memory
         * cannot be used unless the readonly memory kvm capability is present. */
        fprintf(stderr, "qemu: pflash with kvm requires KVM readonly memory support\n");
        exit(1);
    }

    pc_system_flash_init(rom_memory);
}

If we use pflash to flash UEFI bios, the sysfw_firmware_init() still call old_pc_system_rom_init(), the right operation is to call pc_system_flash_init() for pflash. So the cuurent pc_system_flash_init() is not useful.

Please @sameo @rbradford help review this patch, thanks!

### DIFF
--- a/hw/i386/pc.c
+++ b/hw/i386/pc.c
@@ -852,7 +852,7 @@ void pc_memory_init(PCMachineState *pcms,
     }
 
     /* Initialize PC system firmware */
-    sysfw_firmware_init(rom_memory, !pcmc->pci_enabled);
+    sysfw_firmware_init(rom_memory, !pcmc->pci_enabled, true);
 
     option_rom_mr = g_malloc(sizeof(*option_rom_mr));
     memory_region_init_ram(option_rom_mr, NULL, "pc.rom", PC_ROM_SIZE,
--- a/hw/i386/sysfw.c
+++ b/hw/i386/sysfw.c
@@ -102,7 +102,7 @@ static void pc_isa_bios_init(MemoryRegion *rom_memory,
  * it is passed to pc_isa_bios_init(). Merging several drives for isa-bios is
  * not supported.
  */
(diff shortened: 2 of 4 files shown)


yangzhon:
Robs' commit (bc9ed74), 

This make bios memory region read/write, if the change like this, pc_system_flash_init(rom_memory) is not useful, do we need remove CONFIG_PFLASH_CFI01=y in virt target mk file ?

rbradford (changes requested):
Not a reflection on your patch but I don't want to merge this until we have qboot / kernel direct boot testing integrated into NATS.

sameo:
@yangzhon We realized that at least seabios needs that region and we are going to support seabios as well. I think we should close that PR.

yangzhon:
@sameo ,

Yes, i will close this PR, thanks!

PR #159, opened 2018-10-25
Support MSI with ACPI GED device

sboeuf:
This pull request tries to move forward with the goal for NEMU to remove all unneeded legacy component. In this case, it moves the GED device away from its dependency on the IOAPIC, by generating interrupts through MSI message model instead.

It assumes the `evged` driver has been modified accordingly by adding the support for a new ACPI method: `_MSI`. This method allows the guest OS to pass down to NEMU the information about where to send the MSI message, and the vector value.

Based on those assumptions, this pull request defines this new `_MSI` method as part of the `GED` device, and it creates some extra IO ports in order to retrieve the information written by the guest OS by evaluating the `_MSI` method.
Once the MSI info are retrieved, any CPU/Memory/Device hotplug will result in sending an MSI message directly to the virtual APIC, which bypasses the need for the IOAPIC.


include/hw/acpi/ged.h
```diff
@@ -26,18 +26,29 @@
 
 #define ACPI_GED_EVENT_IO_BASE 0xb000
```
mcastelino, line 27: @sboeuf unrelated by can this move to virt.h. Do we need it here? This should be chosen by the platform.

include/hw/acpi/ged.h
```diff
@@ -26,18 +26,29 @@
...
+#define ACPI_GED_IRQ_SEL_LEN     0x4
+#define ACPI_GED_IRQ_SEL_INIT    0x0
+#define ACPI_GED_IRQ_SEL_CPU     0x1
+#define ACPI_GED_IRQ_SEL_MEM     0x2
+#define ACPI_GED_IRQ_SEL_NVDIMM  0x4
+#define ACPI_GED_IRQ_SEL_PCI     0x8
+#define ACPI_GED_IRQ_REG_LEN     ACPI_GED_IRQ_SEL_LEN
+#define ACPI_GED_MSI_OFFSET      (ACPI_GED_IRQ_SEL_OFFSET + ACPI_GED_IRQ_REG_LEN)
+#define ACPI_GED_MSI_IDX_OFFSET  0x4
+#define ACPI_GED_MSI_ADDR_OFFSET 0x8
```
mcastelino: @sboeuf The MSI addr register is actually 64 bit. I know that the upper part is constant currently. But we should still allow it to be programmed back to QEMU. This keep the logic future proof.

hw/acpi/ged.c
```diff
@@ -182,14 +182,38 @@ void build_ged_aml(Aml *table, const char *name, uint32_t ged_irq,
...
                aml_int(ACPI_GED_EVENT_IO_BASE + ACPI_GED_IRQ_SEL_OFFSET),
-               ACPI_GED_IRQ_SEL_LEN));
+               ACPI_GED_IRQ_REG_LEN));
+    /* Append new field IREG */
     field = aml_field(AML_GED_IRQ_REG, AML_DWORD_ACC, AML_NOLOCK,
-                      AML_WRITE_AS_ZEROS);
-    aml_append(field, aml_named_field(AML_GED_IRQ_SEL,
-                                      ACPI_GED_IRQ_SEL_LEN * 8));
+                          AML_WRITE_AS_ZEROS);
+    {
```
mcastelino, line 218: Why do we have the extra braces here.

hw/acpi/ged.c
```diff
@@ -111,7 +111,7 @@ void build_ged_aml(Aml *table, const char *name, uint32_t ged_irq,
                    GedEvent *events, uint32_t events_size)
 {
     Aml *crs = aml_resource_template();
-    Aml *evt, *field;
+    Aml *msi, *evt, *field;
```
mcastelino, line 131: @sboeuf would be good to add in the commit message that this new method is not part of the ACPI spec.

include/hw/acpi/ged.h
```diff
@@ -68,6 +68,9 @@ typedef struct GEDState {
     MemoryRegion io;
     uint32_t     sel;
     uint32_t     irq;
+    uint32_t     msi_idx;
+    uint32_t     msi_addr;
+    uint32_t     msi_data;
```
mcastelino, line 75: @sboeuf as mentioned in an earlier commit let us make accommodation for handling writes to the full 64 bit of the MSI address register. The actual method can choose to only write to the lower 32 bits.

hw/acpi/ged.c
```diff
@@ -84,9 +84,6 @@ void acpi_ged_event(GEDState *ged_st, qemu_irq *irq, uint32_t ged_irq_sel)
     qemu_mutex_lock(&ged_st->lock);
     ged_st->sel |= ged_irq_sel;
     qemu_mutex_unlock(&ged_st->lock);
-
-    /* Trigger the event by sending an interrupt to the guest. */
-    qemu_irq_pulse(irq[ged_st->irq]);
 }
```
mcastelino, line 96: Nice. I like that his got moved out of the generic logic and moved into the platform specific hotplug logic.

hw/acpi/ged.c
```diff
@@ -182,16 +191,49 @@ void build_ged_aml(Aml *table, const char *name, uint32_t ged_irq,
...
+        aml_append(field, aml_named_field(AML_GED_MSI_ADDR,
+                                          ACPI_GED_MSI_ENTRY_LEN * 8));
+        /* Append new entry MDAT */
+        aml_append(field, aml_named_field(AML_GED_MSI_DATA,
+                                          ACPI_GED_MSI_ENTRY_LEN * 8));
+    }
     aml_append(dev, field);
 
+    /* Append _MSI method to pass down the MSI parameters from the guest OS */
+    msi = aml_method("_MSI", 3, AML_SERIALIZED);
```
mcastelino: It will be good to document here the relationship between the declaration of this method and the GSI values specified in the GED _CRS. When _MSI is present the GSI value is not used.

mcastelino (commented):
Three main observations. But it has come together very nicely.

1. Let us move the IO Port address definition to platform specific header file. (virt.h?)
2. Also let us future proof ourselves and provide the ability to use the full 64 bits of MSI Address Register. 
3. Also it will be document the relationship between _GSI and the _CRS interrupt resources in the commit messages or code. (As PR descriptions do not survive in the actual git repo)

include/hw/acpi/ged.h
sboeuf: Well it is provided by the platform, but we suggest a default value from the `ged.h`. Is that okay?

include/hw/acpi/ged.h
sboeuf: Yeah let me add an ADRL and ADRH to make the diff between low and high parts of the address.

hw/acpi/ged.c
sboeuf: It's nothing functional, just a matter of making the code more readable.

hw/acpi/ged.c
sboeuf: Will do!

include/hw/acpi/ged.h
sboeuf: Yes I'll rework that.

hw/acpi/ged.c
sboeuf: Yeah as we discussed it makes more sense.

hw/acpi/ged.c
sboeuf: No, the need for the GSI value is independent from MSI, it depends on the multiplexing.
Let's say we have some platform using GED with no multiplexing, they still want the EVT method to be called with the right `Arg0`. Otherwise, EVT won't be able to differentiate which notification method should be called.

include/hw/acpi/ged.h
sboeuf: done!

include/hw/acpi/ged.h
sboeuf: done!

sboeuf:
@mcastelino 
Here is what I've updated on this PR:
>Let us move the IO Port address definition to platform specific header file. (virt.h?)

Well I had previously defined a default address from ged.h, but the value used is provided by the platform. I didn't change anything yet, let me know what you think here.

>let us future proof ourselves and provide the ability to use the full 64 bits of MSI Address Register.

I have implemented 4 parameters instead of 3, this way we can carry both `addr_lo` and `addr_hi`.

>Also it will be document the relationship between _GSI and the _CRS interrupt resources in the commit messages or code. (As PR descriptions do not survive in the actual git repo)

I have created a new commit adding exclusively some extra comments, and I have written a detailed commit message to summarize MSI support for GED.

Please let me know if you have more comments/concerns :)

hw/acpi/aml-build.c
```diff
@@ -865,6 +865,31 @@ Aml *aml_interrupt(AmlConsumerAndProducer con_and_pro,
     return var;
 }
 
+/*
+ * Interrupt MSI Descriptor
+ * Type 1, Large Item Name 0x13
```
liujing2, line 870: Just curious why we choose 0x13? I read spec that 0xf~0x7f are all reserved.

hw/acpi/aml-build.c
```diff
@@ -865,6 +865,31 @@ Aml *aml_interrupt(AmlConsumerAndProducer con_and_pro,
...
+                       uint64_t msi_id, uint8_t vect_count)
+{
+    Aml *var = aml_alloc();
+    uint8_t irq_flags = con_and_pro | (level_and_edge << 1)
+                        | (high_and_low << 2) | (shared << 3);
+    uint16_t len = 10;
+
+    assert(vect_count > 0);
+
+    build_append_byte(var->buf, 0x93); /* MSI irq descriptor */
```
liujing2, line 884: So 0x93 is the value you choose for InterruptMSI as a large resource type that needs to be added to ACPI spec right?

liujing2 (commented):
Hi @sboeuf I think gsi in nemu part can be removed. Not sure because I didn't do fully test for now. Just read the codes. PR here to do some full tests https://github.com/sboeuf/nemu/pull/1
Otherwise,  this looks good to me!

rbradford:

PR #162, opened 2018-10-30
nats: Add support for direct kernel boot with OVMF

rbradford:
This change extends NATS to add support for specifying the boot
mechanism (direct or bootloader) and adds supports for booting the Clear
Linux kernel with the corresponding image using OVMF as the BIOS.

Signed-off-by: Rob Bradford <robert.bradford@intel.com>

PR #163, opened 2018-10-30
nats: Add support for controlling the boot process

rbradford:
To the distro struct, add a slice of bootConfigs which can be used to
control the boot including BIOS path and optionally kernel path and
kernel command line.

The contents of this struct is then used when constructing the QEMU
command line.

Further add support for direct kernel boot when using Clear Linux.

### DIFF
--- a/tools/CI/nats/nemu_amd64_test.go
+++ b/tools/CI/nats/nemu_amd64_test.go
@@ -16,6 +16,11 @@ import (
 	"github.com/intel/govmm/qemu"
 )
 
+const (
+	clearKernelFileName    = "org.clearlinux.kvm.4.18.16-293"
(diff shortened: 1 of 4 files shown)


rbradford:
@sboeuf  thank you for looking at #162! However i've refactored things slightly in this version and I think it's a bit better. WDYT?

rbradford:
Build time increases from 23 minutes to 28 minutes. Seems reasonable to me?

sboeuf (approved):
@rbradford the refactoring looks fine to me!

About the time it takes to run the whole CI, 5 minutes is a pretty big bump, even if it's obviously worth it.
Could we change Jenkins so that we try to parallelize as much as possible, by using parallel jobs?

rbradford:
@sboeuf I opened #164 to look into this

PR #176, opened 2018-11-13
[DNM] nats: Debug virt failures related to entropy

sboeuf:


rbradford:
@sboeuf There have been a lot of good builds on this. Let's run this for another 24-36 hours. If it continues to stay green then I think we might have a solution to our SSH flakiness that we see on Ubuntu. In preparation for that do you think you could tweak the commit message and drop the change to the (deprecated) start_qemu.sh script.

rbradford:
Superseded by #181

PR #177, opened 2018-11-15
Revert "nats: Temporarily disable CPU hotplug test on virt"

rbradford:
This reverts commit d4a2edcd1b4ca61e6c56c73d328e88a0b59da1d6.


sboeuf:
@rbradford did you find the root cause?

rbradford:
I've hooked this into a periodic build to run it quite a bit over the next 24 hours. I can't explain why this test is now passing on the CI and on a VM where yesterday the same test was almost always failing.

rbradford:
Out of the 98 builds of this PR, only 4 failed and those failure could be unrelated. Therefore i'm merging this change in and re-enabling the test.

PR #179, opened 2018-11-19
sysfw: Bump BIOS region size

rbradford:
The default seabios image is greater than 128KiB so ensure that the
mapping can handle it.

### DIFF
--- a/hw/i386/sysfw.c
+++ b/hw/i386/sysfw.c
@@ -222,7 +222,7 @@ static void old_pc_system_rom_init(MemoryRegion *rom_memory, bool isapc_ram_fw)
     g_free(filename);
 
     /* map the last 128KB of the BIOS in ISA space */
-    isa_bios_size = MIN(bios_size, 128 * KiB);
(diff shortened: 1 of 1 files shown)


sboeuf (commented):
@rbradford Looks good, but I don't understand how we ended up with 128 here. I thought that `pc` and `q35` were still working with SeaBIOS?

rbradford:
@sboeuf It's quite easy to build a seabios image that fits into 128KiB vs 256KiB. So the default kinda makes sense. The default configuration however is larger than 128KiB and so only some of the code is mapped into the region just below 1MiB where it jumps to after leaving real mode. I've been able to build a seabios image  that is 128KiB so i'm less worried about this change but I think it does no harm (although we might need to make sure any e820 tables cover the larger size....)

rbradford:
@sameo i'd like to merge this as the default seabios config is a 256KiB rom size. And although we'll produce an optimised config/build I feel like we should support the default configuration.

sameo (approved):

PR #183, opened 2018-11-22
Fix ACPI HW-reduced sleep (shutdown support)

rbradford:
The incorrectly reported \_S5 object was pushing the kernel to using EFI for shutdown which is its last resort. Once that object was fixed it then generated a kernel panic due to the 0 memory address used for the sleep status register.

With this change shutdown works on seabios on the virt platform.


sameo (approved):
One typo in the c2dfd2f61b377c878d8b3ad387fac320c8b05502 commit message: s/teh/the/.
Otherwise LGTM.

PR #184, opened 2018-11-22
nats: Re-enable Reboot test for Clear Linux

rbradford:
This test was flaky in the past but we've made several NATS related
changes recently which may mean this test is no longer flaky.


rbradford:
Will add a periodic build job for this PR.

sameo (approved):

Issue #129, opened 2018-10-09
SMBIOS for virt 


liujing2:
@sameo @rbradford @yangzhon @mcastelino 
Two key points in conclustion.
1. Do we need open CONFIG_SMBIOS for virt platform in order to use DMI?
1. If yes, there're 2 sub issues.
    1. CONFIG_SMBIOS=y is set in default-configs/x86_64-softmmu.mak, but actually when hw/i386/fw.c fw_build_smbios() check it, it doesn't show as expect. So Nemu doesn't build smbios tables for guest and dmesg shows [    0.000000] DMI not present or invalid.
    1. After fixing i by workaround, only q35 guest can show correct DMI dmesg, virt still shows [    0.000000] DMI not present or invalid.

liujing2:
Some answers as below.

1. @mcastelino mentioned some reasons for DMI in IRC.
```
<mrcastel> the next issue we need to figure out is why DMI info is not passed to Linux for virt
<mrcastel> DMI info is sent in properly by QEMU/ovmf to Linux for q35 and pc
<mrcastel> we have made some mistake in QEMU or ovmf that prevents it
<mrcastel> THe DMI issue is that the kernel says
<mrcastel> [    0.000000] DMI not present or invalid
<mrcastel> for the virt platform
<mrcastel> the DMI not present is there for both virt and Q35 with nemu
<mrcastel> but in normal qemu and ovmf the DMI information is present
<mrcastel> we need to figure out what code change we did caused the DMI info to disappear
<mrcastel> we need the DMI tables to be setup for segments to work
<mrcastel> if you boot mainline QEMU and OVMF you will see DMI present with Q35
<mrcastel> with our nemu we do not see it with Q35 or virt
<mrcastel> we need to figure out why
<mrcastel> so we broke DMI support somewhere
```

And mails mentioned by @mcastelino 
>We will need a simplified code just to report minimal DMI. Just platform name and date. If not the Linux kernel assumes that it is on a really old bios and disables fine features. 

>Let me look at the minimal requirements we need to report. 


2. If we need open CONFIG_SMBIOS, two sub-issues.
For the first one, @yangzhon has root caused as follows.
>As for DMI in Q35 machine type in NEMU repo, this issue is clear now.

> I added CONFIG_SMBIOS in fw_build_smbios in commit id(84d425aee)

> The file hw/i386/fw.c is obj build, but the previous smbios is common-obj build
>ifeq ($(CONFIG_SMBIOS),y)
[...]

PR #188, opened 2018-11-26
Cleanup of multiple segment work

liujing2:
Hi @sameo @rbradford @mcastelino ,
This is the cleanup of multiple segment. For now, GED notify() all the segment when hotplugging. It is a little tricky to direct fetch the segment number that users want to hotplug in ged operation, because ged doesn't know it. I am designing a \_SB method which does not belongs to any segment, and let ged call it. This is WIP. 
Thanks!
Jing


liujing2:
@mcastelino @sameo I updated the GED changes as a separate patch now. So it can only notify the specified segment to do hotplug/unplug. 
To make your review convenient, let me copy the test commands as follows.
```
sudo /home/liujing/work/code/nemu/x86_64_virt-softmmu/qemu-system-x86_64_virt \
-bios /home/liujing/work/code/edk2/Build/OvmfX64/DEBUG_GCC48/FV/OVMF.fd \
        -nographic -nodefaults -L . \
        -machine virt,accel=kvm,kernel_irqchip -cpu host -m size=512M,slots=3,maxmem=4G \
        -smp sockets=1,cpus=4,cores=2,maxcpus=8 \
        -device pci-virt,id=s1.pcie.0,segment-nr=1 \
        -device pci-virt,id=s2.pcie.0,segment-nr=2 \
        -device pci-virt,id=s3.pcie.0,segment-nr=3 \
        -device virtio-blk-pci,drive=image,bus=s1.pcie.0 -drive if=none,id=image,file=/home/liujing/work/image/clear-21950-kvm.img \
        -device virtio-serial-pci,id=virtio-serial0,bus=s1.pcie.0 -device virtconsole,chardev=charconsole0,id=console0 -chardev stdio,id=charconsole0 \
        -monitor telnet:127.0.0.1:5000,server,nowait \
        -qmp unix:hypervisor.sock,server,nowait \
        -kernel /home/liujing/work/code/linux-stable/arch/x86/boot/bzImage \
        -append 'root=/dev/vda3 rw rootfstype=ext4 data=ordered rcupdate.rcu_expedited=1 pci=lastbus=0 pci=realloc=on tsc=reliable no_timer_check reboot=t noapictimer console=hvc0 iommu=off panic=1 movablecore=1G loglevel=8 console=ttyS0' \
        -device sysbus-debugcon,iobase=0x402,chardev=serialcon -chardev file,path=/tmp/serial-log,id=serialcon \

```

hotplug command:
```
(qemu)  device_add virtio-net-pci,id=n1,bus=s2.pcie.0
```

hw/i386/virt/memory.c
```diff
@@ -25,7 +25,7 @@
 #include "qemu/error-report.h"
 #include "cpu.h"
 
-#define VIRT_LOWMEM 0x80000000
+#define VIRT_LOWMEM 0x40000000
```
sboeuf, line 28: It'd be nice to add a comment here in the code explaining why the low mem is 1G.

hw/acpi/pcihp.c
```diff
@@ -332,6 +332,10 @@ void acpi_pcihp_init(Object *owner, AcpiPciHpState *s, PCIBus *root_bus,
                           "acpi-pci-hotplug", s->io_len);
     memory_region_add_subregion(address_space_io, s->io_base, &s->io);
 
+    /* Register property only one time */
+    if (object_property_get_qobject(owner, ACPI_PCIHP_IO_BASE_PROP, NULL))
```
sboeuf: Well I guess it depends if we're sharing the IO regions across all segments, but are we?

Maybe I'm misunderstanding here, but in case we have multiple segments, are we going to have multiple IO ports defined?

include/hw/i386/virt.h
```diff
@@ -66,13 +66,13 @@ typedef struct {
 #define VIRT_ACPI_GED_IRQ 0x10
 
 #define PCI_LITE_PCIEXBAR_BASE  0x40000000
-#define PCI_LITE_PCIEXBAR_SIZE  (0x10000000) /* 256M for bus 0 */
+#define PCI_LITE_PCIEXBAR_SIZE  (0x100000) /* 1M for only one bus */
```
sboeuf, line 69: Do we want to do that? I feel like we won't be able to support devices requiring lot of memory if we do so.

include/hw/i386/virt.h
```diff
@@ -66,13 +66,13 @@ typedef struct {
...
 
 #define PCI_LITE_HOLE_START_BASE 0x90000000
 #define PCI_LITE_HOLE64_START_BASE 0x100000000ULL
 
 #define PCI_VIRT_PCIEXBAR_BASE  0x50000000
-#define PCI_VIRT_PCIEXBAR_SIZE  0x10000000
+#define PCI_VIRT_PCIEXBAR_SIZE  0x100000
```
sboeuf, line 75: Same comment here, and I would add that I thought the additional segments should be over 4G so that we can use as much memory as we want.

hw/pci-host/pci_virt.c
```diff
@@ -187,6 +189,8 @@ static PCIHostState *pci_virt_init(DeviceState *dev,
     pcie_host_mmcfg_update(pcie, 1, mcfg_base, PCI_VIRT_PCIEXBAR_SIZE);
 
     e820_add_entry(mcfg_base, PCI_VIRT_PCIEXBAR_SIZE, E820_RESERVED);
+    //fw_cfg->entries[0][0x21].data ==> e820_table
```
sboeuf: A comment in the code summarizing your commit message would be great ;)

hw/acpi/pcihp.c
```diff
@@ -87,13 +87,13 @@ static void *acpi_set_bsel(PCIBus *bus, void *opaque)
 
 static void acpi_set_pci_info(AcpiPciHpState *s)
 {
-    static bool bsel_is_set;
     unsigned bsel_alloc = ACPI_PCIHP_BSEL_DEFAULT;
 
-    if (bsel_is_set) {
+    /* To avoid setting the same property again */
+    if (s->root && object_property_get_qobject(OBJECT(s->root),
+                                         ACPI_PCIHP_PROP_BSEL, NULL)) {
```
sboeuf, line 94: Nice!

hw/pci-host/pci_virt.c
```diff
@@ -0,0 +1,287 @@
...
+{
+    PCIVirtHost *pci_virt = PCI_VIRT_HOST(host_bridge);
+
+    snprintf(pci_virt->bus_path, sizeof(pci_virt->bus_path),
+             "%04x:00", pci_virt->segment_nr);
+    return pci_virt->bus_path;
+}
+
+static Property pci_virt_props[] = {
+    DEFINE_PROP_UINT16("segment-nr", PCIVirtHost, segment_nr, 1),
```
sboeuf, line 253: I think `segment` would be better than `segment-nr`. And if you use `bus` or `addr` as example, there's no `-nr` suffix.

hw/pci-host/pci_lite.c
```diff
@@ -179,6 +180,24 @@ static void pci_lite_realize(DeviceState *dev, Error **errp)
     }
 }
 
+/*
+ * Pass MCFG base and PCI hole base for i386/virt.
```
sboeuf, line 184: Couldn't we use a single file to pass all those information?

Or maybe even reuse some of the existing files by injecting those new values that we need?

sboeuf (commented):
A first review with a few questions, but I'll do a second pass later.

hw/acpi/pcihp.c
liujing2: @sboeuf Thanks for your review! This property actually has nothing with our virt platform. It's only used in pc platform and I am thinking to separate it in a new function or something else...

hw/i386/virt/memory.c
liujing2: Yes, this is because of [DMI issue result](https://github.com/intel/nemu/issues/129#issuecomment-429321621) And so I put all the mcfg below 4GB.

hw/acpi/pcihp.c
sboeuf: Oh okay, so splitting into a new function would be a good idea. Something like `acpi_pcihp_set_properties(AcpiPciHpState *s)` that would be only called from `pc`.

liujing2: OK, will do that and update.

hw/pci-host/pci_virt.c
liujing2: Oh yes, this commit need be squash into previous one later so this would be deleted :)

hw/pci-host/pci_virt.c
liujing2: I agree! Will change that.

hw/pci-host/pci_lite.c
liujing2: Ummm, I am not sure. Manohar  suggested using one file per segment to mark some related information... I am also confusing which way to choose.

include/hw/i386/virt.h
liujing2: @sboeuf  So how about 1M only for MCFG size not for pci hole size. So large bar can be covered as original qemu status.

liujing2:
Mark: We need specify -device xxx,bus=pcie.0 if we want use pcie.0 (segment 0) as the bus, because once other segment/root-bus is added, qemu uses QLIST_INSERT_HEAD so other new added segment will be default to find for non-specified device.
```
static void pci_host_bus_register(DeviceState *host)
{
    PCIHostState *host_bridge = PCI_HOST_BRIDGE(host);

    QLIST_INSERT_HEAD(&pci_host_bridges, host_bridge, next);
}

```

PR #191, opened 2018-11-28
nats: Enable qboot

rbradford:
Add Clear Linux testing with qboot.


rbradford:
@sameo yup networking doesn't work with qboot, i'll look into that.

PR #194, opened 2018-11-30
build: Disable gnutls support

rbradford:
This was being autodetected and leading to build issues.

Fixes: #198


rbradford:
This change addresses all the dynamically enabled config options that might break us.

sameo (approved):

PR #199, opened 2018-12-18
Cleanup of multiple segment work

liujing2:
@mcastelino  @sboeuf  @sameo @rbradford  I update the codes based on 3.1 and look forward to any comments. Thanks!

BTW ovmf changes here https://github.com/intel/ovmf-virt/pull/4


liujing2:
Mark: We need specify -device xxx,bus=pcie.0 if we want use pcie.0 (segment 0) as the bus, because once other segment/root-bus is added, qemu uses QLIST_INSERT_HEAD so other new added segment will be default to find for non-specified device.
```
static void pci_host_bus_register(DeviceState *host)
{
    PCIHostState *host_bridge = PCI_HOST_BRIDGE(host);

    QLIST_INSERT_HEAD(&pci_host_bridges, host_bridge, next);
}
```

liujing2:
Usage now:
 -device pci-virt,id=s1.pcie.0,max-bus=1 \
Limitation and assumption:

    *Guest kernel needs some DMI information to allow MCFG above 4GB.
     So for not, MCFG allocation is just below 4GB to adapt a non-modified guest.

    *MCFG space is contiguous for all segments below 32 bit.

    *Shrink the low memory size to 1GB to reserve more space for multiple segment.
     The MCFG is reserved as below.
     0x40000000~0x4fffffff: MCFG for segment 0, maximum bus in theory supported
     0x50000000~0xbfffffff: MCFG for more segment, limit max-bus=1 for each.

    *Dynamically getting the hole64 though marking the pci hole64 address
     used by pci-lite host bridge.

    *Do not allocate pci hole below 32 bit for pci-virt host bridge. For devices
     that need 32 bit hole, please plug to segment zero.

    *pci-lite host bridge (segment zero) has pci hole below 32 bit as,
     0xc0000000~IO_APIC_DEFAULT_ADDRESS - 1

    *Add a pci-virt host bridge as follows, notes max-bus needs to be 1 for now,
     for the convenience of firmware.
     -device pci-virt,id=s1.pcie.0,max-bus=1

liujing2:
Qemu 3.1 upstream fixed the bug that: if the root-bus you hotplugged onto doesn't support pci-hole32, while your device needs that, the hotplug interrupt from ged will only be received once. That is to say, the second interrupt of hotplug/remove will not be received by guest evged driver. The first hotplug will be "success" by lspci.
For 3.1, guest driver will always receive the interrupt, but the hotplug onto a non-supported root bus will fail and lspci show nothing even for the first hotplug.

rbradford:

PR #208, opened 2019-03-11
virtio-pci: Correctly expose vector count for virtio-rng-pci

rbradford:
The vector count should be  1 + the number of virtqueues which is one
for virtio-rng. The other PCI devices correctly define this property and
default value. This will then enable the use of MSI-X which prevents the
kernel from falling back to legacy interrupts which we do not support.

### DIFF
--- a/hw/virtio/virtio-pci.c
+++ b/hw/virtio/virtio-pci.c
@@ -2489,6 +2489,11 @@ static void virtio_rng_pci_realize(VirtIOPCIProxy *vpci_dev, Error **errp)
                              NULL);
 }
 
+static Property virtio_rng_pci_properties[] = {
+    DEFINE_PROP_UINT32("vectors", VirtIOPCIProxy, nvectors, 2),
+    DEFINE_PROP_END_OF_LIST(),
+};
+
 static void virtio_rng_pci_class_init(ObjectClass *klass, void *data)
 {
     DeviceClass *dc = DEVICE_CLASS(klass);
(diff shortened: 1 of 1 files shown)


hw/virtio/virtio-pci.c
```diff
@@ -2489,6 +2489,11 @@ static void virtio_rng_pci_realize(VirtIOPCIProxy *vpci_dev, Error **errp)
                              NULL);
 }
 
+static Property virtio_rng_pci_properties[] = {
+    DEFINE_PROP_UINT32("vectors", VirtIOPCIProxy, nvectors, 2),
```
mcastelino, line 2493: @rbradford so this bug will manifest fully in platforms without significant legacy interrupts? Or will it manifest even in a pc and q35 but with lesser impact?

sboeuf (approved):
@rbradford thanks for spotting this! LGTM

Follow up question: is this bug part of QEMU upstream? I think it'd be worth submitting a patch there if that's the case :)

hw/virtio/virtio-pci.c
sameo: IIUC, the former.

sameo (approved):
LGTM
Thanks a lot @rbradford

hw/virtio/virtio-pci.c
rbradford: @mcastelino On pc and q35 the kernel will fall back to `vp_find_vqs_intx` vs `vp_find_vqs_msix` but that completely

PR #229, opened 2019-05-07
virtio-fs: Add latest patches fixing the output of the virtiofs daemon

sboeuf:
This pull request adds the latest patches coming from RedHat regarding the virtiofs daemon.
They fix the output that is printed by the daemon.


sboeuf:
@rbradford please take a look and merge if ok for you!

rbradford (approved):

